PR #57, opened 2021-10-28
Integrate Filecoin Consensus natively in Eudico

adlrocha:
This PR:
- Adds native support to filecoin consensus in Eudico so it can be run in the same way as the delegated and PoW consensus chains.
- Integrates genesis block generation for fil consensus to up the Shard Actor in the initial state tree. This offers shard support to the filecoin consensus, and we can now run filecoin consensus in the root chain (which wasn't possible with the previous setup).
- Adds basic support of filcns in shards.

Shortcomings:
- We support a single miner with pre-sealed data in genesis. For now I think this is OK, as long as we figure the right way to add new miners in runtime (we should be able to figure this out by "reverse engineering" how is done in Lotus).


adlrocha:
Closing in favor of https://github.com/filecoin-project/eudico/pull/144. We are not planning to support Filecoin Consensus in subnets anymore.

Issue #41, opened 2021-10-11
Checkpointing to Bitcoin (B2) MVP


sa8:
These are the main tasks for the checkpointing project (project B2).

- [ ] Integrate Zondax's work and fix outstanding issues
  - [x] #106
  - [x] #102
  - [ ] #98
  - [x] #103
  - [x] #105
  - [ ] #104
  - [x] #107
  - [ ] #120
- [x] Deploy Eudico testnet
  - [x] #42
  - [x] #46
- [x] Deploy Bitcoin regtest
  - [x] Deploy Bitcoin regtest and run a full node (1 node enough)
  - [x] Write reusable scripts that can be used to replicate the system.
- [ ] Distributed Key Generation
  - [x] #43
  - [x] #100
  - [x] #108
  - [ ] #109
  - [ ] #110
  - [ ] #184
- [ ] Signing
  - [x] #51
  - [ ] #52
  - [x] #44
  - [x] #60
  - [x] #47
  - [x] #111
  - [x] #48
  - [ ] #112
  - [ ] #113
  - [ ] #58
  - [x] (Optional) remove the signature aggregator and automate it using Eudico
- [x] #45
- [x] #55
- [x] Push checkpoint transaction (Taproot+OP_RETURN) to Bitcoin blockchain
  - [x] Push checkpoint transaction (Taproot+OP_RETURN) to Bitcoin blockchain with dummy key
  - [x] Push checkpoint transaction (Taproot+OP_RETURN) to Bitcoin blockchain integrated with DKG
- [x] #117
- [ ] Verification
  - [x] Get checkpoint transaction from the bitcoin blockchain
  - [x] #114
  - [x] Verify checkpoint against data from Eudico chain
  - [ ] #116
- [ ] Measurements and Monitoring
  - [ ] #56
  - [x] #115
- Bugs
  - [ ] #201

PR #83, opened 2021-12-03
Bitcoin Checkpointing MVP

rllola:
## Abstract

This PR add the checkpointing feature in Bitcoin (see #41).

## Details
- [x] Load distributed key file `config.toml` (see `./data/alice/config.toml`)
- [x] Ping bitcoind to see if available
- [x] Connect to participants using Libp2p PubSub
- [x] Generate new distributed keys among participants
- [x] First participant publish the funding transaction to the taproot address (loaded pubkey + block 0 CID)
- [x] Every 20 blocks create a new checkpoint
- [ ] Create a config file in S3
- [x] Collect precedent checkpoint tx (add taproot address to wallet and list transactions)
- [ ] Verify checkpoints
- [ ] Kick misbehaving participant
- [x] Sign using tweaked value in a distributed fashion
- [ ] Mocked power actors

## Run it
You will need docker bitcoind running locally under port `18443` with rpc user `satoshi` and rpc password `amiens`.
Under `data` you will find preconfigure participants in order to quickly start (Alice, Bob and Charlie).
```
$ make eudico
$ ./scripts/taproot.sh
```
It will start 3 nodes and a miner.


adlrocha:
> After rebasing I started to have this error :
2021-12-03T14:23:59.999+0100    ERROR   events  events/events_called.go:352     event diff fn failed: cbor input had wrong number of fields
I am not sure how to fix it.

Try running `make type-gen` to generate all cbor schemas again. It seems like some state that you are unmarshalling the wrong state into an object. Let me know if you don't manage to fix it and I can try to look at it next week.

rllola:
After adding the mocked power it seems that the error has disappeared.

chain/checkpointing/constant.go
```diff
@@ -0,0 +1,7 @@
+package checkpointing
+
+var FEE float64 = 0.01
+var S3_HOST string = "minio.deadbrain.corp"
+var ACCESS_KEY_ID string = ""
+var SECRET_ACCESS_KEY string = ""
+var BUCKET_NAME string = "eudico"
```
adlrocha, line 7: Suggestion for the future: what about adding these in the config? I don't expect these to be constants in every runtime. Each user will probably chose their own, right?

chain/checkpointing/network.go
```diff
@@ -0,0 +1,61 @@
...
+	return c
+}
+
+func (n *Network) Next(ctx context.Context) *protocol.Message {
+	msg, err := n.sub.Next(ctx)
+	if err != nil {
+		panic(err)
+	}
+
+	// Converting a pubsub.Message into a protocol.Message
```
adlrocha, line 33:
```suggestion
	// Unwrapping protocol message from PubSub message
```
Nit-pick to make it more readable. It is not quite transforming it, it is just fetching the protocol message marshalled data from the pubsub envelope.

Feel free to dissregard it if it doesn't make sense.

chain/checkpointing/storage.go
```diff
@@ -0,0 +1,36 @@
...
+	"github.com/minio/minio-go/v7"
+	"github.com/minio/minio-go/v7/pkg/credentials"
+)
+
+func StoreConfig() {
+	// Initialize minio client object.
+	minioClient, err := minio.New(S3_HOST, &minio.Options{
+		Creds:  credentials.NewStaticV4(ACCESS_KEY_ID, SECRET_ACCESS_KEY, ""),
+		Secure: false,
+	})
```
adlrocha, line 18: As mentioned above, it'd be great in the future if we can gather these values from config.

chain/checkpointing/storage.go
```diff
@@ -0,0 +1,36 @@
...
+	if err != nil {
+		panic(err)
+	}
+
+	fmt.Println(minioClient)
```
adlrocha, line 23: Can we use "github.com/ipfs/go-log/v2" logs instead of `fmt.Println`? It is super straightforward to use, let me know if you need an example of how to use it.

chain/checkpointing/sub.go
```diff
@@ -0,0 +1,506 @@
...
+func NewCheckpointSub(
+	mctx helpers.MetricsCtx,
+	lc fx.Lifecycle,
+	host host.Host,
+	pubsub *pubsub.PubSub,
+	api impl.FullNodeAPI,
+) (*CheckpointingSub, error) {
+
+	ctx := helpers.LifecycleCtx(mctx, lc)
+	// Starting shardSub to listen to events in the root chain.
```
adlrocha, line 70:
```suggestion
	// Starting checkpoint listener
```

chain/checkpointing/sub.go
```diff
@@ -0,0 +1,506 @@
...
+	e, err := events.NewEvents(ctx, &api)
+	if err != nil {
+		return nil, err
+	}
+
+	var config *keygen.TaprootConfig
+	// Load configTaproot
+	if _, err := os.Stat(os.Getenv("EUDICO_PATH") + "/share.toml"); errors.Is(err, os.ErrNotExist) {
+		// path/to/whatever does not exist
+		fmt.Println("No share file saved")
```
adlrocha, line 80: Sanity-check: We do nothing and return an empty config if no share is stored here? Don't we need to return an error or do somethign to avoid the config from being `nil`?

If this is the case we could probably change this if statement to make it more legible and do something like:
```go
if (dir doesn't exist) {
   // Generate config
}
```

adlrocha: Oh, I see, if there is no config you just start an empty ShardingSub to be populated in the future. Then you probably don't need the `else` and a single `if' as proposed can do the trick.

chain/checkpointing/sub.go
```diff
@@ -0,0 +1,506 @@
...
+}
+
+func BuildCheckpointingSub(mctx helpers.MetricsCtx, lc fx.Lifecycle, c *CheckpointingSub) {
+	ctx := helpers.LifecycleCtx(mctx, lc)
+
+	// Ping to see if bitcoind is available
+	success := BitcoindPing()
+	if !success {
+		// Should probably not panic here
+		panic("Bitcoin node not available")
```
adlrocha, line 492: We can probably emit a warning saying that checkpointing is not available because there's no connection to a Bitcoin node, but allow the node to start. In this case we should be sure that we don't start any of the listening routines or event listeners, as they may error like crazy.

chain/checkpointing/sub.go
```diff
@@ -0,0 +1,506 @@
...
+
+		idsStrings := c.orderParticipantsList()
+
+		fmt.Println("Participants list :", idsStrings)
+
+		ids := c.formIDSlice(idsStrings)
+
+		id := party.ID(c.host.ID().String())
+
+		threshold := 2
```
adlrocha, line 170: Why is the threshold 2? Are we assuming three nodes here? We should probably adapt this according to the number of miners in the network, right?

chain/checkpointing/sub.go
```diff
@@ -0,0 +1,506 @@
...
+		n := NewNetwork(ids, c.sub, c.topic)
```
adlrocha, line 171: `NewNetwork` is independent in every round of the DKG and that is why we start it every time a new state change is detected? If this is the case we should probably try to find a better name for the `Network` object :) From my first read, it felt like that is the permanent network layer used by the DKG protocol (and reused in every round)

chain/checkpointing/sub.go
```diff
@@ -0,0 +1,506 @@
...
+		f := frost.KeygenTaproot(id, ids, threshold)
+
+		handler, err := protocol.NewMultiHandler(f, []byte{1, 2, 3})
+		if err != nil {
+			fmt.Println(err)
+			panic(err)
```
adlrocha, line 177: Why panic here?

chain/checkpointing/sub.go
```diff
@@ -0,0 +1,506 @@
...
+		}
+		c.LoopHandler(ctx, handler, n)
+		r, err := handler.Result()
+		if err != nil {
+			fmt.Println(err)
+			panic(err)
+		}
+		fmt.Println("Result :", r)
+
+		c.newconfig = r.(*keygen.TaprootConfig)
```
adlrocha, line 187: To avoid panicking here try using something like:
```go
c.newconfig, ok = r.(*keygen.TaprootConfig)
if !ok{
   return xerrors.Errorf("state change propagated is the wrong type")
}
```

data/alice/config.toml
```diff
@@ -0,0 +1,185 @@
+[API]
```
adlrocha, line 1: Just noting it here for the future. Before merging and getting this ready for production, we should probably move these configs to a `scripts/taproot` that we can use for testing purposes, or remove them. Also, we can probably remove all the commented lines if they are already in the config template.

chain/checkpointing/sub.go
```diff
@@ -0,0 +1,506 @@
...
+			if err != nil {
+				panic(err)
+			}
+			data := ts.Cids()[0]
+			err = c.initiate(data.Bytes())
+			if err != nil {
+				panic(err)
+			}
+			return false, nil, nil
+		}
```
adlrocha, line 230: If I understand correctly what we do here is to in the first state change to create a new checkpoint including the keys for the genesis tipset, right?

chain/checkpointing/sub.go
```diff
@@ -0,0 +1,506 @@
...
+
+		// ZONDAX TODO
+		// Activate checkpointing every 20 blocks
+		fmt.Println("Height:", newTs.Height())
+		if newTs.Height()%20 == 0 {
+			fmt.Println("Check point time")
+
+			// Initiation and config should be happening at start
```
adlrocha, line 238: We may need to revise this logic to be sure that:
* The checkpoint is running every 20 blocks.
* We may need to wait a few epochs before detecting a checkpointing epoch to ensure finality (the tipset we detect in the event change may be reverted and not be final, and committing a reverted tipset in a checkpoint may be quite bad).
* We should check that we are only committing the checkpoints one. Due to the way `checkFunc` works, it may trigger the `ChangeFunc` every time the state change is detected, which could end up translating in a checkpoint being committed several times.

Just wanted to note it here so we test it at some point.

adlrocha: Also, we should probably add a config parameter to configure the checkpointing period.

chain/checkpointing/sub.go
```diff
@@ -0,0 +1,506 @@
...
+			if c.init && c.config != nil {
+
+				data := oldTs.Cids()[0]
+
+				var partyList string = ""
+				for _, partyId := range c.orderParticipantsList() {
```
adlrocha, line 244: Out of a curiosity, what is the reason for ordering the list of participants?

chain/checkpointing/sub.go
adlrocha: Re-reading the function again, it may be a good idea to detect the change here, wait for a few epochs, and the trigger the changeFunc to create a new checkpoint. That way we ensure finality. I do something similar here in case you need inspiration: https://github.com/filecoin-project/eudico/blob/d08ea35d0e4bfe437b7d1910829300fbb49a18f0/chain/consensus/hierarchical/subnet/events.go#L203

chain/checkpointing/sub.go
```diff
@@ -0,0 +1,506 @@
...
+	if err != nil {
+		panic(err)
+	}
+
+	/*
+	 * Orchestrate the signing message
+	 */
+
+	f := frost.SignTaprootWithTweak(c.config, ids, hashedTx[:], c.tweakedValue[:])
+	n := NewNetwork(ids, c.sub, c.topic)
```
adlrocha, line 378: We should definitely the change the name of the `Network` struct, because without additional docs it is very misleading.

chain/checkpointing/sub.go
```diff
@@ -0,0 +1,506 @@
...
+	c.sub = sub
+
+	c.listenCheckpointEvents(ctx)
+}
+
+func (c *CheckpointingSub) LoopHandler(ctx context.Context, h protocol.Handler, network *Network) {
+	for {
+		msg, ok := <-h.Listen()
+		if !ok {
+			network.Done()
```
adlrocha, line 293: If you don't need to clean anything, you can probably remove this function and just return when the channel is closed.

chain/checkpointing/sub.go
```diff
@@ -0,0 +1,506 @@
...
+		panic("Bitcoin node not available")
+	}
+
+	fmt.Println("Successfully pinged bitcoind")
+
+	c.Start(ctx)
+
+	lc.Append(fx.Hook{
+		OnStop: func(ctx context.Context) error {
+			// Do we need to stop something here ?
```
adlrocha, line 501: Probably not much. In `Start` you are joining the `keygen` topic, you should probably leave it here,but I can't think of anything else that needs to be done right now.

chain/checkpointing/sub.go
```diff
@@ -0,0 +1,506 @@
...
+
+	idsSlice := party.NewIDSlice(_ids)
+
+	return idsSlice
+}
+
+func (c *CheckpointingSub) prefundTaproot() error {
+	taprootAddress := PubkeyToTapprootAddress(c.pubkey)
+
+	payload := "{\"jsonrpc\": \"1.0\", \"id\":\"wow\", \"method\": \"sendtoaddress\", \"params\": [\"" + taprootAddress + "\", 50]}"
```
adlrocha, line 451: Quick question: why are you defining the raw payload instead of creating a struct and then marshalling? It may be more readable and less prone to bugs.

chain/checkpointing/sub.go
```diff
@@ -0,0 +1,506 @@
...
+	value, scriptPubkeyBytes := GetTxOut(c.ptxid, index)
+
+	if scriptPubkeyBytes[0] != 0x51 {
+		fmt.Println("Wrong txout")
+		index = 1
+		value, scriptPubkeyBytes = GetTxOut(c.ptxid, index)
+	}
+	newValue := value - FEE
+
+	payload := "{\"jsonrpc\": \"1.0\", \"id\":\"wow\", \"method\": \"createrawtransaction\", \"params\": [[{\"txid\":\"" + c.ptxid + "\",\"vout\": " + strconv.Itoa(index) + ", \"sequence\": 4294967295}], [{\"" + newTaprootAddress + "\": \"" + fmt.Sprintf("%.2f", newValue) + "\"}, {\"data\": \"" + hex.EncodeToString(data) + "\"}]]}"
```
adlrocha, line 349: Quick question: why are you defining the raw payload instead of creating a struct and then marshalling? It may be more readable and less prone to bugs.

chain/checkpointing/util.go
```diff
@@ -0,0 +1,288 @@
...
+	return h.Sum(nil)
+}
+
+func Sha256(data []byte) []byte {
+	h := sha256.New()
+	h.Write(data[:])
+	return h.Sum(nil)
+}
+
+func TaprootSignatureHash(tx []byte, utxo []byte, hash_type byte) ([]byte, error) {
```
adlrocha, line 48: @sa8 may I ask for you for some help reviewing this part? I am not super familiar with our design and I may be missing some details. Happy to do it sync together if it works best for you (that way I can learn a bit more about how the protocol works :) ).

chain/checkpointing/util.go
```diff
@@ -0,0 +1,288 @@
...
+
+	taprootAddressLegacy, err := bech32.Encode("bcrt", conv)
+	if err != nil {
+		fmt.Println(err)
+		log.Fatal("Couldn't produce our tapproot address.")
+	}
+	return taprootAddressLegacy
+}
+
+func ApplyTweakToPublicKeyTaproot(public []byte, tweak []byte) []byte {
```
adlrocha, line 133: @sa8 same for the next few function related with taproot if possible :)

chain/checkpointing/util.go
```diff
@@ -0,0 +1,288 @@
...
+		panic("cant retrieve previous transaction")
+	}
+	taprootTxOut := result["result"].(map[string]interface{})
+	scriptPubkey := taprootTxOut["scriptPubKey"].(map[string]interface{})
+	scriptPubkeyBytes, _ := hex.DecodeString(scriptPubkey["hex"].(string))
+
+	return taprootTxOut["value"].(float64), scriptPubkeyBytes
+}
+
+func jsonRPC(payload string) map[string]interface{} {
```
adlrocha, line 252: Why not returning an error in this function?

chain/checkpointing/sub.go
```diff
@@ -0,0 +1,506 @@
...
+		// ZONDAX TODO
+		// Activate checkpointing every 20 blocks
+		fmt.Println("Height:", newTs.Height())
+		if newTs.Height()%20 == 0 {
+			fmt.Println("Check point time")
+
+			// Initiation and config should be happening at start
+			if c.init && c.config != nil {
+
+				data := oldTs.Cids()[0]
```
adlrocha, line 241: Quick question (and I apologize in advance if this is obvious but I lack deep knowledge of the protocol): the data included in the checkpoint is just the tipset keys of the epoch when the checkpoint is being committed? Didn't we also add information about the power table in the checkpoint, or that is implicit in the key used to sign?

chain/checkpointing/util.go
```diff
@@ -0,0 +1,288 @@
...
+}
+
+type TaprootConfigTOML struct {
+	Thershold          int
+	PrivateShare       string
+	PublicKey          string
+	VerificationShares map[string]VerificationShare
+}
+
+func TaggedHash(tag string, datas ...[]byte) []byte {
```
adlrocha, line 30: If all these functions are exclusively used in this package, we can probably make them private.

adlrocha:
@rllola, I did a first pass through the code and left a few comments. Let me know if you feel any part of the code needs a deeper review at this point. Feel free to resolve the comments as you fix them or if you see that they don't apply, and ping me whenever you need another review.

chain/checkpointing/constant.go
rllola: Done!

I have added a special section with those value so they can de loaded at start. It still can be improved.

chain/checkpointing/util.go
rllola: @sa8 Here the BIP 341 detailing how a message should be serialized before being signed (https://github.com/bitcoin/bips/blob/master/bip-0341.mediawiki#common-signature-message). The scheme for Taproot has changed and is a bit more complicated.

You can also look at this python script used to test against the golang function (https://github.com/Zondax/fil-taproot/blob/main/scripts/main.py).

This step is being done right before signing and it return a hash. The hash is the message then signed by the participants.

chain/checkpointing/util.go
rllola: Yes, it should return an error. Fixing this.

chain/checkpointing/sub.go
```diff
@@ -0,0 +1,600 @@
...
+					panic(err)
+				}
+
+				go c.CreateCheckpoint(ctx, data, hash)
+			}
+		}
+
+		// If Power Actors list has changed start DKG
+		// Changes detected so generate new key
+		if oldSt.MinerCount != newSt.MinerCount {
```
sa8, line 318: we may only need to check that the size of symmetrical difference between the two sets is bigger than some threshold u.

chain/checkpointing/util.go
```diff
@@ -0,0 +1,285 @@
...
+		// Particular case where we are already in the process of adding the key
+		// because we are using 1 bitcoin node for all
+		return true
+	}
+
+	return false
+}
+
+func GetTaprootScript(pubkey []byte) string {
+	return "5120" + hex.EncodeToString(pubkey)
```
sa8, line 186: The "5120" is not clear to me here

rllola: It is the taproot script. The taproot script should start with 1 and is followed by the pubkey.

chain/checkpointing/sub.go
```diff
@@ -0,0 +1,600 @@
...
+			fmt.Println("Generate new config")
+
+			return true, nil, nil
+		}
+
+		return false, nil, nil
+	}
+
+	err := c.events.StateChanged(checkFunc, changeHandler, revertHandler, 5, 76587687658765876, match)
```
sa8, line 327: what does "76587687658765876" refers to?

adlrocha: This is the confidence threshold used to determine if the `StateChangeHandler` should be triggered. It is an absurdly high number so the metric used to determine if to trigger it or not is the number of tipsets that have passed in the heaviest chain (the `5` you see there)

chain/checkpointing/sub.go
adlrocha: Solved. @sa8 clarified it to me. It is used to choose deterministically the miners that need to sign.

data/alice/config.toml
rllola: We agreed to move everything that is under `data` will be move in Minio and downloaded before starting the demo. Those are pre-generated keys and config only required when developing/running the demo.

chain/checkpointing/sub.go
rllola: This part has been now removed. The initialization is being done outside of eudico.

chain/checkpointing/sub.go
rllola: I will add a comment. Any name suggestions ? I re-use `Network` because it is how it was called in the lib and our script.

chain/checkpointing/sub.go
rllola: This part of code is outdated. But yes we also have the list of particpants who signed the checkpoint. The latest state of the PR reflect this.

chain/checkpointing/sub.go
adlrocha: `NewRound` maybe? Or something like `NewState`? Let's try to find something more self-describing than `Network`  if we can. It led me to confusion the first time I read it :)

rllola: I thought of `NewRound` too but it could be confusing at the cryptographic level because there is several rounds of communication when you do the DKG/signing... What about `NewNetInterface` ? Because it is an interface for communication between participants. They can receive and send messages through this instance.

What about `Communication` ? `NewCommunicationRound` ?

PR #85, opened 2021-12-07
Fund/Release methods in Subnets

adlrocha:
Implement methods to move funds in/out of a subnet through the parent chain. For this we implemented:
- [x] A custom Reward Actor that implements an `ExternalFunding` function to "mint" new FIL in the subnet's address where funds are being sent from the parent chain.
- [x] A new `CrossMsgs` field to Filecoin blocks that can be used to include cross-msgs to be validated as part of the consensus  in subnets. This is used to enforce the consistency and ordering of cross messages entering the subnet before they are applied and they trigger a state change.
- [x] `Fund` and `Release` methods in SCA to move funds in and out of the subnet. 
  - [x] `Fund` freezes the funds for an address in the parent chain and notifies the subnet to mint new FIL.
  - [x] `Release` burns in the subnet the FILs that want to be released for an address in the parent chain. This method sends a transaction to the `BurntAddress` of the subnet with the corresponding FILs, and adds information in the checkpoint to the parent to notify the SCA actor that the FILs for the address can be released.
    - [x] As part of this protocol checkpoints have been modified to include metadata about the cross-messages being propagated through the hierarchy (you'll find new structures like `CrossMsgMeta` that helps us with this task). 
- [x] As part of the PR, some common parts of the consensus have been refactored and decoupled from specific consensus implementations: see `chain/consensus/common`.

----
Deferred for the next PR so this one doesn't keep growing infinitely:
- [ ] Include a process in miners that asks the subnet manager for cross messages pending for validation/application in the subnet. The subnet manager needs to be subscribed to changes in the state of the parent and root SCA actor to apply top-down and down-top messages conveniently.
- [ ] Add validation checks for cross-messages. Miners in a subnet only pick up the Cid for a message (or a set of messages) directed to the subnet. It'll need to inspect if the messages and MsgMeta behind those cross-messages have already been pushed by the sending subnet, or it needs to be proactively fetched in order to check build the new block with cross messages (Welcome to the data availability problem!)
- [ ] Create new `TipsetExecutor` for subnets that runs an `ApplyImplicitMessage` in subnet consensus algorithms to mint the new FIL in a specific address when funds are frozen in the parent. The RewardActor is the one responsible for "funding new addresses", and is triggered through the `TipSetExecutor`'s ImplicitMessages in `ApplyBlocks` as it is currently done for miner rewards. When peers detect that funds from an address have been frozen in the parent chain, an `ExternalFund` ImplicitMessage is triggered by the subnet consensus algorithm in every peer (this ensures state change consistency for every peer --again, analogous to how miner rewards work). 
   - [ ] The application of certain cross messages (like `Release` down-top messages) may require the implementation of new specific function in SCA (for instance, `ApplyReleaseMsg`). 
- [ ] CLI commands to sends funds and release them from a subnet.
 
All of this builds the foundation for cross-chain transactions. `Fund` and  `Release` are the core primitives used to trigger cross-chain transactions propagated through checkpoints.


adlrocha:
As @vukolic noted in Slack, this design is incomplete, as we don't have a way to sync the application of funding `ImplicitMessages` between miners in the subnet.

> Example: parent subnet is at tipset height P.65 and saying we need to mint in child subnet A. However, one miner in a child subnet A is A.100 when it delivers P.65 and the other miner in subnet A is at A.102 at the moment it delivers P.65

adlrocha:
We are converging to the following improvement to the original protocol to ensure every funding message is synced in the same epoch for all miners in the subnet (and to allow miners to sync the chain from scratch):
* Every funding message in SCA of the parent chain is assigned an incremental nonce.
* Miners in the consensus algorithm of the subnet check the nonce (n) of the latest funding message validated in the last block of the subnet. When proposing a new block, miners include funding messages from the parent chain with nonce > n (if any)
* The consensus algorithm runs in the subnet usual, but additional checks are added for funding messages (to double-check that the messages included are > n and that they are committed in the parent chain).
* When a new block is validated in the subnet, all funding messages included in the block are applied.

This also fixes an additional problem that I found out while thinking about using checkpoints for synchronization, or applying messages exclusively when the block is executed, which is the synchronization of new miners. If we exclusively depend on state change info, there is no easy way for new miners to sync the subnets chain from scratch and apply funding message in the right epoch.

Things to bear in mind: 
- This impacts latency (of ability to spend) a bit (at most one block).
- This deterministic order will only work for top-down transactions (but it is OK, we use checkpoints bottom-up).
- Throughout the implementation it would be useful to have a latency and gas costs end-to-end analysis of the solution.

PR #88, opened 2021-12-20
Validate and apply cross messages in subnet consensus

adlrocha:
This PR is a continuation of https://github.com/filecoin-project/eudico/pull/85. Here we include all the things we left behind in the previous PR:

- [x] `ApplyMsg` function in SCA used to trigger its execution through `ApplyImplicitMessage` in `compute_state` of the subnet consensus when a cross-message is verified by the subnet consensus.
- [x] Include a process in miners that asks the subnet manager for cross messages pending for validation/application in the subnet when miners propose new blocks.
- [x] Add validation checks for cross-messages. Miners in a subnet only pick up the Cid for a message (or a set of messages) directed to the subnet. It'll need to inspect if the messages and MsgMeta behind those cross-messages have already been pushed by the sending subnet, or it needs to be proactively fetched in order to check build the new block with cross messages (Welcome to the data availability problem!)
- [x] Create new `TipsetExecutor` for subnets that runs an `ApplyImplicitMessage` in subnet consensus algorithms to mint the new FIL in a specific address when funds are frozen in the parent. The RewardActor is the one responsible for "funding new addresses", and is triggered through the `TipSetExecutor`'s ImplicitMessages in `ApplyBlocks` as it is currently done for miner rewards. When peers detect that funds from an address have been frozen in the parent chain, an `ExternalFund` ImplicitMessage is triggered by the subnet consensus algorithm in every peer (this ensures state change consistency for every peer --again, analogous to how miner rewards work). 
   - [ ] The application of certain cross messages (like `Release` down-top messages) may require the implementation of new specific function in SCA (for instance, `ApplyReleaseMsg`). 
- [x] CLI commands to sends funds and release them from a subnet.


chain/consensus/common/crossmsg.go
```diff
@@ -0,0 +1,180 @@
...
+	case hierarchical.Release:
+		panic("Not implemented")
+	case hierarchical.Cross:
+		panic("Not implemented")
+	}
+
+	return xerrors.Errorf("Unknown cross-msg type")
+}
+
+// checkTopDownMsg validates the topdown message.
```
LefKok, line 41: TopDown and DownTop is confusing, maybe TopDown and BottomUp?

LefKok: is top down from parent to child? then why can't the root send messages to the children?

chain/consensus/common/crossmsg.go
```diff
@@ -0,0 +1,180 @@
...
+// - It checks that the msg nonce is larger than AppliedDownTopNonce in the subnet SCA
+// Recall that applying crossMessages increases the AppliedNonce of the SCA in the subnet
+// where the message is applied.
+// - It checks that the cross-msg is committed in the sca of the parent chain
+func checkTopDownMsg(pstore blockadt.Store, parentSCA, snSCA *sca.SCAState, msg *types.Message) error {
+	// Check valid nonce in subnet where message is applied.
+	if msg.Nonce < snSCA.AppliedDownTopNonce {
+		return xerrors.Errorf("topDown msg nonce below AppliedDownTop nonce in subnet where is applied")
```
LefKok, line 49: Error message a bit confuding? Maybe "msg nonce reuse"?

chain/consensus/common/crossmsg.go
```diff
@@ -0,0 +1,180 @@
...
+
+	return xerrors.Errorf("Unknown cross-msg type")
+}
+
+// checkTopDownMsg validates the topdown message.
+// - It checks that the msg nonce is larger than AppliedDownTopNonce in the subnet SCA
+// Recall that applying crossMessages increases the AppliedNonce of the SCA in the subnet
+// where the message is applied.
+// - It checks that the cross-msg is committed in the sca of the parent chain
+func checkTopDownMsg(pstore blockadt.Store, parentSCA, snSCA *sca.SCAState, msg *types.Message) error {
```
LefKok, line 46: who calls this? shouldn't it only be called if the messages is committed in the parent? why do we need to recheck in L53

chain/consensus/common/crossmsg.go
```diff
@@ -0,0 +1,180 @@
...
+	}
+
+	if ret.ExitCode != 0 {
+		return xerrors.Errorf("reward application message failed (exit %d): %s", ret.ExitCode, ret.ActorErr)
+	}
+	fmt.Println(">>>>>>>>>>> IMPLICIT FUND MESSAGE APPLIED", apply.Cid(), ret)
+	return nil
+}
+
+func getSCAState(ctx context.Context, sm *stmgr.StateManager, submgr subnet.SubnetMgr, id hierarchical.SubnetID) (*sca.SCAState, blockadt.Store, error) {
```
LefKok, line 150: Does this need checking? I don't get much 😅

chain/consensus/hierarchical/subnet/manager/crossmsg.go
```diff
@@ -0,0 +1,307 @@
...
+
+type crossMsgPool struct {
+	lk   sync.RWMutex
+	pool map[hierarchical.SubnetID]*lastApplied
+}
+
+type lastApplied struct {
+	lk      sync.RWMutex
+	topdown map[uint64]abi.ChainEpoch // nonce[epoch]
+	downtop map[uint64]abi.ChainEpoch // nonce[epoch]
```
LefKok, line 44: do we need a nonce for messages coming from child to parent? what if there are multiple children do they need coordinate?

chain/consensus/hierarchical/subnet/manager/crossmsg.go
```diff
@@ -0,0 +1,307 @@
...
+	height  abi.ChainEpoch
+}
+
+func (cm *crossMsgPool) getPool(id hierarchical.SubnetID, height abi.ChainEpoch) *lastApplied {
```
LefKok, line 48: I'd expect different pools for every relationship (one for parent, one each for children)

chain/consensus/common/crossmsg.go
adlrocha: > TopDown and DownTop is confusing, maybe TopDown and BottomUp?

Great point. Will do a global renaming.

> Is top down from parent to child? then why can't the root send messages to the children?

Yes. Some transactions may not need to go all the way up to the root, addresses may not have the available balance in the root, and the SCA in the parent is the one tracking the available supply in the subnet. There may be a better way of handling all of this. I'll add all of this to the spec and we can discuss it further.

chain/consensus/common/crossmsg.go
adlrocha: Agree. Thanks!

chain/consensus/common/crossmsg.go
adlrocha: This is just picking up the current state in the SCA of the subnet. IIUC there is no checking going on here.

chain/consensus/hierarchical/subnet/manager/crossmsg.go
adlrocha: I don't think this is needed. The transactions picked up in the pool have already been committed and checked by the SCA. The SCA aggregates and commits all transactions from childs and assigns a unique nonce to all of them. This pool picks up all of these transactions and apply them in order in the corresponding network consensus.

This is also related with you comment above. In any case, I definitely get the question because this PR is missing the code for the SCA. I really need to update the spec to have all the design in the same place. I'll ping you there when it is ready. Thanks!

adlrocha:
Implemented `Fund` subnet. Deferring `Release` function for future PRs to avoid this PR from becoming too long. It may be a good to revisit some of the `TODO:` in the consensus validations once bottom-up cross-msg application are also implemented.

PR #94, opened 2022-01-13
Implement Release bottom-up cross msg

adlrocha:
- [x] Implement `Release` cmd
- [x] Validation of bottom-up in subnet consensus
- [x] Content resolution protocol for checkpoints and CrosMsgMeta
  - [x] Push/Pull approaches
- [x] New `f04` address and moved `SubnetID` to [fork of `go-address`](https://github.com/adlrocha/go-address).
  - `f04` addresses include subnet context information in them. These addresses are compatible with hierarchical-consensus cross-messages
- [x] Modified `list-subnets` and `list-checkpoitns` to include information about  circulating supply and `cross-msgs` propagated, respectively. 
- [x] General refactor to include readibility and code-org.   
- [x] Removed mining rewards in subnets. Miners are only granted gas rewards.


chain/consensus/hierarchical/actors/sca/sca_apply.go
```diff
@@ -41,6 +52,67 @@ func applyFund(rt runtime.Runtime, msg types.Message) {
 	}
 }
 
+func applyRelease(rt runtime.Runtime, msg types.Message) {
```
adlrocha, line 55: @Lefkok, if you could have a "sanity-check look" to this piece of code I would really appreciate it. This function applies `release` msgs (i.e. bottom-up cross-messages that release funds from a subnet to the parent chain).

(The PR is huge, so don't worry about reviewing it fully, many are just refactors and cosmetic changes).

Thanks

chain/consensus/hierarchical/subnet/resolver/resolver.go
```diff
@@ -0,0 +1,499 @@
+package resolver
```
adlrocha, line 1: @Lefkok, and this is the implementation of the cross-messages resolution protocol (the one that pull-push cross-messages from other subnets). In case you want to also look at it (this is less of a priority than the previous comment).

chain/consensus/hierarchical/actors/sca/sca_apply.go
```diff
@@ -41,6 +52,67 @@ func applyFund(rt runtime.Runtime, msg types.Message) {
...
+	// Release funds to the destination address.
+	code := rt.Send(rto, builtin.MethodSend, nil, msg.Value, &builtin.Discard{})
+	if !code.IsSuccess() {
+		noop(rt, code)
+	}
+}
+
+func bottomUpStateTransition(rt runtime.Runtime, st *SCAState, msg types.Message) {
+	// Bottom-up messages include the nonce of their message meta. Several messages
+	// will include the same nonce. They need to be applied in order of nonce.
```
LefKok, line 95: the same nonce? this can cause arbitrary re-ordering of these messages breaking determinism

chain/consensus/hierarchical/actors/sca/sca_apply.go
```diff
@@ -41,6 +52,67 @@ func applyFund(rt runtime.Runtime, msg types.Message) {
...
+
+	// As soon as we see a message with the next msgMeta nonce, we increment the nonce
+	// and start accepting the one for the next nonce.
+	// FIXME: Once we have the end-to-end flow of cross-message this REALLY needs to
+	// be revisited.
+	if st.AppliedBottomUpNonce+1 == msg.Nonce {
```
LefKok, line 101: shouldn't we first make sure the transition happened and then increase the nonce? Otherwise we can get 3 messages that start in order but finish out of order (assuming this code can be called from multiple threads)

chain/consensus/hierarchical/actors/sca/sca_apply.go
```diff
@@ -41,6 +52,67 @@ func applyFund(rt runtime.Runtime, msg types.Message) {
...
+		// Increment latest nonce applied for bottomup
+		incrementNonce(rt, &st.AppliedBottomUpNonce)
+	}
+
+	// NOTE: Check if the nonce of the message being applied is the subsequent one (we could relax a bit this
+	// requirement, but it would mean that we need to determine how we want to handle gaps, and messages
+	// being validated out-of-order).
+	if st.AppliedBottomUpNonce != msg.Nonce {
```
LefKok, line 109: Can a multi-threaded execution increase the nonce to i+2 in L103 before this code is checked for message with nonce i+1?

chain/consensus/hierarchical/actors/sca/sca_apply.go
```diff
@@ -41,6 +52,67 @@ func applyFund(rt runtime.Runtime, msg types.Message) {
...
+		rt.Abortf(exitcode.ErrIllegalState, "the message being applied doesn't hold the subsequent nonce (nonce=%d, applied=%d)",
```
LefKok, line 110: What happens in case of Abort? Can i get back my burned funds?

LefKok: or retry with a higher nonce

chain/consensus/hierarchical/actors/sca/sca_apply.go
```diff
@@ -41,6 +52,67 @@ func applyFund(rt runtime.Runtime, msg types.Message) {
...
+			rt.Abortf(exitcode.ErrIllegalState, "subnet for actor hasn't been registered yet")
+		}
+		// Even if the actor has balance, we shouldn't allow releasing more than
+		// the current circulating supply, as it would mean that we are
+		// releasing funds from the collateral.
+		if sh.CircSupply.LessThan(msg.Value) {
+			rt.Abortf(exitcode.ErrIllegalState, "wtf! we can't release funds below the circulating supply. Something went wrong!")
+		}
+		sh.CircSupply = big.Sub(sh.CircSupply, msg.Value)
+		st.flushSubnet(rt, sh)
```
LefKok, line 83: can the flush fail asynchronously? We might need to block and wait to make sure the substraction has taken effect before releasing the funds

chain/consensus/hierarchical/actors/sca/sca_apply.go
adlrocha: We use the same nonce for cross-msgs include in the same `CrossMsgMeta` to identify that they belong to the "same package". This ensures the ordering between cross-msg propagated by several child subnets. However, before being applied, the original nonce of the message is recovered (through the cross-msg resolution protocol) and all messages are sorted conveniently. Thus, the order of application is deterministic, and someone trying to forge it would reach an inconsistent state.

That being said, the comment is really on point, the code that handles message sorting is outside the actor (it is [here](https://github.com/filecoin-project/eudico/pull/94/files#diff-96caa44c6fa84d9d0cf1e6b9820105e89227da2112d9ed8653727dfc6af5718dR272), in case you want to have a quick look at it)

chain/consensus/hierarchical/actors/sca/sca_apply.go
adlrocha: Not really. This code belongs to the actor and can only be triggered sending a message to it, so it is run sequentially "one message at a time".

chain/consensus/hierarchical/actors/sca/sca_apply.go
adlrocha: That is a great question. The answer in the current implementation is _no_. This is something that we need to figure out how we want to do it. We could maybe add some information in checkpoint to notify back the source that certain message could not be applied (although this may be quite slow). I think we briefly mentioned this in our previous sync, but let me add a point to discuss it again in our next sync.

adlrocha:
@LefKok, thank you for you comments, they've been really useful to catch a few minor bugs. I am reasonably happy with the current state of the code after your review. Let me know if you see anything blocking or if you want to have an additional pass to something, if not I'll skim one last time through the code and merge it by EOD. Thanks.

PR #95, opened 2022-01-13
Fixing cross-msg compatibility

adlrocha:
In hierarchical consensus we've introduced a new type of message in blocks, `CrossMsgs`. In order to be able to sync with a chain that started before hierarchical consensus, we need to handle [`CompactedMessages`](https://github.com/filecoin-project/eudico/blob/4a3ccfed70a6a920a10fe43ca57dcf7f4785b31c/chain/exchange/protocol.go#L167-L168) and [blocks](https://github.com/filecoin-project/eudico/blob/4a3ccfed70a6a920a10fe43ca57dcf7f4785b31c/chain/types/blockmsg.go#L13) and after including cross messages.

This PR solves the backward compatibility but breaks the `delegated` and `tspow` consensus, as well as subnet consensus. validations. As this won't be a priority until we reach a final design of hierarchical consensus and we move into production I defer to then the investigation of what may be failing and the fix (we may change things in the future and realize there are better ways to reach backward compatibility).

### DIFF
--- a/chain/consensus/common/cns_validations.go
+++ b/chain/consensus/common/cns_validations.go
@@ -334,11 +334,12 @@ func checkBlockMessages(ctx context.Context, str *store.ChainStore, sm *stmgr.St
 		return err
 	}
 
-	mrcid, err := tmpstore.Put(ctx, &types.MsgMeta{
-		BlsMessages:   bmroot,
-		SecpkMessages: smroot,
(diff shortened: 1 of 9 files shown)


adlrocha:
Close in favor of https://github.com/filecoin-project/eudico/issues/218

PR #123, opened 2022-02-10
added FAQ

sa8:
I added an FAQ file that contains some questions asked to @adlrocha and their answers.
I'm hoping others can contribute too and can use this file when they have a question.

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -1,0 +1,32 @@
+# Eudico FAQ
+
+**Q: How to import a wallet?**
+
+**A:** Generate a key:
+```
+./lotus-keygen -t spec256k1
+```
+Import it (as default if needed):
+```
+./eudico wallet import –-as-default –-format=json-lotus $file
+```
+
+**Q: How to generate the cbor files when creating a new actor?**
+
+**A:** The Cbor library is used to help with marshalling and unmarshalling.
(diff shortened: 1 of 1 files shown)


FAQ.md
```diff
@@ -0,0 +1,32 @@
+# Eudico FAQ
+
```
adlrocha, line 2: What about adding here a note of the like of: "you can probably find many of these commands in lotus docs [link], but in this FAQ you'll find a list of some that may come handy for the typical user flows needed to interact with Eudico".

FAQ.md
```diff
@@ -0,0 +1,32 @@
...
+```
+func (t *$YourStruct) MarshalCBOR(w io.Writer) error { return nil }
+
+func (t *$YourStruct) MarshalCBOR(w io.Writer) error { return nil }
+```
+Run  `go run gen.go`. 
+
+
+**Q: How do you send a transaction directly from the code?**
+**A:** By using the `api-MpoolPushMessage.` See it use [here](https://github.com/filecoin-project/eudico/blob/113829e7fc115daac08ea0217170baddcb7788ba/chain/consensus/hierarchical/subnet/manager/manager.go#L375-L391).
```
adlrocha, line 29:
```suggestion
**A:** By using the `api-MpoolPushMessage.` See how to use it [here](https://github.com/filecoin-project/eudico/blob/113829e7fc115daac08ea0217170baddcb7788ba/chain/consensus/hierarchical/subnet/manager/manager.go#L375-L391).
```

adlrocha (approved):
@dnkolegov, would you like to include your notes in this PR? If not, @sa8, feel free to merge and we'll keep iterating.

dnkolegov:
I will add later

PR #124, opened 2022-02-13
fixed bug in resolver when rejoining subnet

adlrocha:
Fix bug for which pubsub topic validator of resolver was not being unregistered when leaving or closing a subnet. This PR fixes the issue.

### DIFF
--- a/chain/consensus/hierarchical/subnet/manager/manager.go
+++ b/chain/consensus/hierarchical/subnet/manager/manager.go
@@ -318,7 +318,8 @@ func (s *SubnetMgr) Close(ctx context.Context) error {
 			return err
 		}
 	}
-	return nil
+	// Close resolver
+	return s.r.Close()
 }
 
 func BuildSubnetMgr(mctx helpers.MetricsCtx, lc fx.Lifecycle, s *SubnetMgr) {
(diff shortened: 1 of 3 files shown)


adlrocha:
@dnkolegov, this is a minor fix, but pinging you just for sanity-check.

chain/consensus/hierarchical/subnet/manager/manager.go
```diff
@@ -318,7 +318,8 @@ func (s *SubnetMgr) Close(ctx context.Context) error {
 			return err
 		}
 	}
-	return nil
+	// Close resolver
+	return s.r.Close()
```
dnkolegov, line 322: Why do we return if an error happened? don't we need to continue?

chain/consensus/hierarchical/subnet/manager/subnet.go
```diff
@@ -167,6 +167,8 @@ func (sh *Subnet) Close(ctx context.Context) error {
 	sh.syncer.Stop()
 	// Close message pool
 	err5 := sh.mpool.Close()
+	// Close resolver.
+	err6 := sh.r.Close()
```
dnkolegov, line 171: I believe err1, err2, ... are not idiomatic names for errors. Let's add TODO?

dnkolegov (approved):

chain/consensus/hierarchical/subnet/manager/subnet.go
adlrocha: Good point, adding TODO.

PR #128, opened 2022-02-15
Support for arbitrary params in cross-net messages

adlrocha:
This PR includes a `SendWithSerializedParams` in the runtime interface (and the vm implementation) to support forwarding cross-net messages with already serialized params to the right destination.

_Note: this required a fork of `filecoin-project/specs-actors` with the new runtime interface._


adlrocha:
@dnkolegov, this should also be a quick one. Thanks

dnkolegov (approved):

PR #129, opened 2022-02-18
Fix README and FAQ

matejpavlovic:
Parts of the README and FAQ were improperly copy-pasted from their
Lotus counterparts and the commands were not working.
Now a new user should be able to run a dev network on the local
machine by following the guide in the README.


adlrocha (approved):
LGTM. Will merge and we can iterate further in the future. Thanks!

PR #133, opened 2022-02-22
Atomic actor primitives. Sample atomic execution actor. Off-chain execution of locked state

adlrocha:
- [x] Actor primitives to run atomic execution protocol.
- [x] Sample implementation of an actor using atomic primitives to perform atomic execution.
- [x] Primitives to perform off-chain execution of locked state.

Next:
- PR1: Implementation of SCA logic to handle atomic execution protocol in parent.
- PR2: Atomic execution manager and CLI cmds to perform atomic executions from Eudico node.

PR #137, opened 2022-03-02
SCA logic for atomic execution in parent chain.

adlrocha:
Depends on: https://github.com/filecoin-project/eudico/pull/133

This PR implements all the logic for the atomic execution protocol in the SCA:
- [x] Initialize new  atomic execution
- [x] Submit output/abort execution


chain/consensus/actors/atomic-replace/gen/gen.go
```diff
@@ -0,0 +1,17 @@
+package main
+
+import (
+	replace "github.com/filecoin-project/lotus/chain/consensus/actors/atomic-replace"
+	gen "github.com/whyrusleeping/cbor-gen"
```
dnkolegov, line 5: non-critical: unsorted packages

chain/consensus/actors/atomic-replace/replace.go
```diff
@@ -0,0 +1,264 @@
...
+import (
+	"github.com/filecoin-project/go-address"
+	"github.com/filecoin-project/go-state-types/abi"
+	"github.com/filecoin-project/go-state-types/cbor"
+	"github.com/filecoin-project/go-state-types/exitcode"
+	actor "github.com/filecoin-project/lotus/chain/consensus/actors"
+	"github.com/filecoin-project/lotus/chain/consensus/hierarchical/atomic"
+	"github.com/filecoin-project/specs-actors/v7/actors/builtin"
+	"github.com/filecoin-project/specs-actors/v7/actors/runtime"
+	cid "github.com/ipfs/go-cid"
```
dnkolegov, line 16: non-critical: unsorted

dnkolegov, line 16: non-critical: unsorted

chain/consensus/actors/atomic-replace/replace.go
```diff
@@ -0,0 +1,264 @@
...
+// where the whole map is locked for an atomic execution.
+// We could use a more fine-grained approach, but this actor
+// has illustrative purposes for now. Consider improving it in
+// future iterations.
+type ReplaceState struct {
+	Owners *atomic.LockedState
+}
+
+type Owners struct {
+	M map[string]cid.Cid
```
dnkolegov, line 39: I believe that `M` is not a good name for public data types. It would work if `M` was used in a seminal paper, etc.
For example, `_, ok := o.M[k]` from below looks strange

chain/consensus/actors/atomic-replace/replace.go
```diff
@@ -0,0 +1,264 @@
...
+}
+
+// Merge implement the merge strategy to follow for our lockable state
```
dnkolegov, line 42: misprint - implements

chain/consensus/actors/atomic-replace/replace.go
```diff
@@ -0,0 +1,264 @@
...
+		if ok {
+			return xerrors.Errorf("merge conflict. key for owner already set")
+		}
+		o.M[k] = v
+	}
+	return nil
+
+}
+
+func ConstructState() (*ReplaceState, error) {
```
dnkolegov, line 61: It is not so obvious from a naming perspective why Owner's `ConstructState` returns `*ReplaceState`

chain/consensus/actors/atomic-replace/replace.go
```diff
@@ -0,0 +1,264 @@
...
+	rt.ValidateImmediateCallerIs(builtin.SystemActorAddr)
+	var st ReplaceState
+	rt.StateTransaction(&st, func() {
+		merge := &Owners{}
+		err := atomic.UnwrapMergeParams(params, merge)
+		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "error unwrapping output from mergeParams")
+		st.merge(rt, merge)
+	})
+
+	return nil
```
dnkolegov, line 195: Is that correct in a sense of returning `*abi.EmptyValue` ?

chain/consensus/actors/atomic-replace/replace.go
```diff
@@ -0,0 +1,264 @@
...
+	// Only system actor can trigger this function.
+	rt.ValidateImmediateCallerIs(builtin.SystemActorAddr)
+
+	var st ReplaceState
+	rt.StateTransaction(&st, func() {
+		output := &Owners{}
+		err := atomic.UnwrapUnlockParams(params, output)
+		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "error unwrapping output from unlockParams")
+		switch params.Params.Method {
+		case 5:
```
dnkolegov, line 208: What is `5`? Can we create a constant or something similar?

chain/consensus/actors/atomic-replace/replace.go
```diff
@@ -0,0 +1,264 @@
...
+func (a ReplaceActor) Abort(rt runtime.Runtime, params *atomic.LockParams) *abi.EmptyValue {
+	// FIXME: We should check here that the only one allowed to abort an execuetion is
+	// the rt.Caller() that locked the state? Or is the system.actor because is triggered
+	// through a top-down transaction?
+	rt.ValidateImmediateCallerAcceptAny()
+
+	var st ReplaceState
+	rt.StateTransaction(&st, func() {
+		switch params.Method {
+		case 5:
```
dnkolegov, line 236: the same comment as above

chain/consensus/actors/atomic-replace/replace.go
```diff
@@ -0,0 +1,264 @@
...
+
+func ValidateLockedState(rt runtime.Runtime, st *ReplaceState) {
+	builtin.RequireNoErr(rt,
+		atomic.ValidateIfLocked([]*atomic.LockedState{st.Owners}...),
+		exitcode.ErrIllegalState, "state locked")
+}
+
+// UnwrapOwners is a convenient function to handle the locked state from the actor.
+func (st *ReplaceState) UnwrapOwners() (*Owners, error) {
+	own := &Owners{}
```
dnkolegov, line 254: Doesn't this look more readable?

```
        var own Owners
	if err := atomic.UnwrapLockableState(st.Owners, &own); err != nil {
		return nil, err
	}
	return &own, nil
```

The pattern is used in other places: e.g., line 220

chain/consensus/actors/atomic-replace/replace.go
```diff
@@ -0,0 +1,264 @@
...
+	var st ReplaceState
+	rt.StateTransaction(&st, func() {
+		ValidateLockedState(rt, &st)
+		own, err := st.UnwrapOwners()
+		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "error unwrapping lockable state")
+		_, ok := own.M[rt.Caller().String()]
+		if ok {
+			rt.Abortf(exitcode.ErrIllegalState, "address already owning something")
+		}
+		own.M[rt.Caller().String()], err = abi.CidBuilder.Sum([]byte(params.Seed))
```
dnkolegov, line 124: what if params is nil?

chain/consensus/actors/atomic-replace/replace.go
```diff
@@ -0,0 +1,264 @@
...
+func (a ReplaceActor) Replace(rt runtime.Runtime, params *ReplaceParams) *abi.EmptyValue {
+	rt.ValidateImmediateCallerAcceptAny()
+
+	var st ReplaceState
+	rt.StateTransaction(&st, func() {
+		ValidateLockedState(rt, &st)
+		own, err := st.UnwrapOwners()
+		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "error unwrapping lockable state")
+		_, ok1 := own.M[rt.Caller().String()]
+		_, ok2 := own.M[params.Addr.String()]
```
dnkolegov, line 145: what if params is nil?

chain/consensus/actors/atomic-replace/replace.go
```diff
@@ -0,0 +1,264 @@
...
+}
+
+func (a ReplaceActor) Unlock(rt runtime.Runtime, params *atomic.UnlockParams) *abi.EmptyValue {
+	// Only system actor can trigger this function.
+	rt.ValidateImmediateCallerIs(builtin.SystemActorAddr)
+
+	var st ReplaceState
+	rt.StateTransaction(&st, func() {
+		output := &Owners{}
+		err := atomic.UnwrapUnlockParams(params, output)
```
dnkolegov, line 205: what if params is nil? should we check it here?

chain/consensus/actors/atomic-replace/replace.go
```diff
@@ -0,0 +1,264 @@
...
+
+func (a ReplaceActor) Abort(rt runtime.Runtime, params *atomic.LockParams) *abi.EmptyValue {
+	// FIXME: We should check here that the only one allowed to abort an execuetion is
+	// the rt.Caller() that locked the state? Or is the system.actor because is triggered
+	// through a top-down transaction?
+	rt.ValidateImmediateCallerAcceptAny()
+
+	var st ReplaceState
+	rt.StateTransaction(&st, func() {
+		switch params.Method {
```
dnkolegov, line 235: The same question: what if params is nil?

chain/consensus/actors/atomic-replace/replace.go
```diff
@@ -0,0 +1,264 @@
...
+			rt.Abortf(exitcode.ErrIllegalArgument, "this method has nothing to unlock")
+		}
+	})
+
+	return nil
+}
+
+func ValidateLockedState(rt runtime.Runtime, st *ReplaceState) {
+	builtin.RequireNoErr(rt,
+		atomic.ValidateIfLocked([]*atomic.LockedState{st.Owners}...),
```
dnkolegov, line 248: Could st be nil?

chain/consensus/hierarchical/actors/sca/sca_actor.go
```diff
@@ -634,3 +639,108 @@ func requireSuccessWithNoop(rt runtime.Runtime, msg types.Message, code exitcode
...
+		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "error getting exec map")
+		// Check if execution already exists.
+		_, found, err := getAtomicExec(execMap, c)
+		builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "error getting exec map")
+		if found {
+			rt.Abortf(exitcode.ErrIllegalArgument, "execution with cid %s already initialized", c)
+		}
+
+		// sanity-check: verify that all messages have same method and are directed to the same actor.
+		method := params.Msgs[0].Method
```
dnkolegov, line 663: Params is not checked before accessing its fields

chain/consensus/actors/atomic-replace/replace.go
adlrocha: In this case, as the package name is `/ipfs/go-cid` the linter suggests (it actually forces in my case) to be after `/filecoin-project/*` dependencies.

adlrocha: Same as above.

chain/consensus/actors/atomic-replace/replace.go
adlrocha: Yes. Actually `*abi.EmptyValue` just replaces `nil` by `struct{}{}` under the hood to avoid marshalling errors.

chain/consensus/actors/atomic-replace/replace.go
adlrocha: Oh, good catch. I created the constant but forgot to update it there. Thanks!

chain/consensus/actors/atomic-replace/replace.go
dnkolegov: sure. I meant grouping them into several groups as in other files.

```
         cid "github.com/ipfs/go-cid"

         "github.com/filecoin-project/lotus/chain/consensus/hierarchical/atomic"
	"github.com/filecoin-project/specs-actors/v7/actors/builtin"
	"github.com/filecoin-project/specs-actors/v7/actors/runtime"
	
```

chain/consensus/actors/atomic-replace/replace.go
adlrocha: I don't think params can ever be nil. Actor methods receive their params after unmarshalling a stream. It would fail somewhere before this can be nil. That being said, I agree that having a sanity-check in critical actors (which is not currently the case) may not be a bad idea.

chain/consensus/actors/atomic-replace/replace.go
adlrocha: Similar to `params`, `st` can't be `nil` as it is the result of unmarshalling the actor state from the stateStore. Actor state is initialized in actors' constructor, so having this would mean that something very wrong happened.

chain/consensus/hierarchical/actors/sca/sca_actor.go
adlrocha: Good call. We should check here if `Msgs` or `LockedStates` are empty. Thanks!

chain/consensus/actors/atomic-replace/replace.go
adlrocha: I guess `ConstructState` is a "de-facto standard" at this point. Is the main function that constructs the state of the actor in construction. In this case the state just happens to have a single data structure for owners.

chain/consensus/actors/atomic-replace/replace.go
adlrocha: Oh, gotcha. Thanks for the clarification.

chain/consensus/actors/atomic-replace/replace.go
dnkolegov: ok. I agree

adlrocha:
@dnkolegov, sorry for the mess. I wanted to rebase with the previous PR and I started getting infinite conflicts out of the blue. It should be fixed by now and ready to merge. Let me know if there is anything else you want me to check. Thanks!

dnkolegov (approved):

PR #138, opened 2022-03-04
Commands to manage atomic executions between subnets

adlrocha:
This PR:
- [x] Includes `subnet atomic` command group to orchestrate and run atomic executions. 
- [x] Includes new commands to deploy new actors in subnets and compute the hierarchical address for a raw address in a subnet.
- [x] Includes support to request locked states in the subnet's content resolution protocol.
- [x] Includes improvements in usability of the locking primitives

### DIFF
--- a/api/api_hierarchical.go
+++ b/api/api_hierarchical.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/filecoin-project/go-address"
 	"github.com/filecoin-project/go-state-types/abi"
+	"github.com/filecoin-project/lotus/chain/consensus/hierarchical/actors/sca"
 	"github.com/filecoin-project/lotus/chain/consensus/hierarchical/checkpoints/schema"
 	"github.com/filecoin-project/lotus/chain/types"
 	"github.com/ipfs/go-cid"
@@ -26,4 +27,8 @@ type HierarchicalCns interface {
 	CrossMsgResolve(ctx context.Context, id address.SubnetID, c cid.Cid, from address.SubnetID) ([]types.Message, error)   // perm:read
 	LockState(ctx context.Context, wallet address.Address, actor address.Address, subnet address.SubnetID,
 		method abi.MethodNum) (cid.Cid, error) // perm:write
+	InitAtomicExec(ctx context.Context, wallet address.Address, inputs map[string]sca.LockedState, msgs []types.Message) (cid.Cid, error) // perm:write
+	ListAtomicExecs(ctx context.Context, id address.SubnetID, addr address.Address) ([]*sca.AtomicExec, error)                            // perm:read
+	ComputeAndSubmitExec(ctx context.Context, wallet address.Address, id address.SubnetID, execID cid.Cid) (sca.ExecStatus, error)        // perm:write
+	AbortAtomicExec(ctx context.Context, wallet address.Address, id address.SubnetID, execID cid.Cid) (sca.ExecStatus, error)             // perm:write
 }
--- a/api/api_hierarchical.go
+++ b/api/api_hierarchical.go
@@ -24,4 +24,6 @@ type HierarchicalCns interface {
 	FundSubnet(ctx context.Context, wallet address.Address, id address.SubnetID, value abi.TokenAmount) (cid.Cid, error)   // perm:write
 	ReleaseFunds(ctx context.Context, wallet address.Address, id address.SubnetID, value abi.TokenAmount) (cid.Cid, error) // perm:write
 	CrossMsgResolve(ctx context.Context, id address.SubnetID, c cid.Cid, from address.SubnetID) ([]types.Message, error)   // perm:read
+	LockState(ctx context.Context, wallet address.Address, actor address.Address, subnet address.SubnetID,
+		method abi.MethodNum) (cid.Cid, error) // perm:write
 }
--- a/api/api_hierarchical.go
+++ b/api/api_hierarchical.go
@@ -27,8 +27,9 @@ type HierarchicalCns interface {
 	CrossMsgResolve(ctx context.Context, id address.SubnetID, c cid.Cid, from address.SubnetID) ([]types.Message, error)   // perm:read
 	LockState(ctx context.Context, wallet address.Address, actor address.Address, subnet address.SubnetID,
 		method abi.MethodNum) (cid.Cid, error) // perm:write
+	UnlockState(ctx context.Context, wallet address.Address, actor address.Address, subnet address.SubnetID, method abi.MethodNum) error  // perm:write
 	InitAtomicExec(ctx context.Context, wallet address.Address, inputs map[string]sca.LockedState, msgs []types.Message) (cid.Cid, error) // perm:write
-	ListAtomicExecs(ctx context.Context, id address.SubnetID, addr address.Address) ([]*sca.AtomicExec, error)                            // perm:read
+	ListAtomicExecs(ctx context.Context, id address.SubnetID, addr address.Address) ([]sca.AtomicExec, error)                             // perm:read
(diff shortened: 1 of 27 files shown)


chain/consensus/hierarchical/subnet/resolver/resolver.go
```diff
@@ -454,19 +551,69 @@ func (r *Resolver) ResolveCrossMsgs(ctx context.Context, c cid.Cid, from address
 
 }
 
+func (r *Resolver) WaitLockedStateResolved(ctx context.Context, c cid.Cid, from address.SubnetID, actor address.Address) chan error {
+	out := make(chan error)
```
dnkolegov, line 555: We should use a buffered channel. This allows the goroutine to exit if we don't collect the error.
```out := make(chan error, 1)```

File: chain/consensus/hierarchical/subnet/resolver/resolver.go [hunk 2]
```diff
@@ -346,24 +362,104 @@ func (r *Resolver) processResponse(ctx context.Context, rmsg *ResolveMsg) (pubsu
...
+	if err != nil {
+		return err
+	}
+	m := &ResolveMsg{
+		Type:   ResponseLocked,
+		From:   r.netName,
+		Cid:    c,
+		Locked: lstate,
+	}
+	return r.publishMsg(m, to)
```
dnkolegov, line 395: I would recommend
```
m := ResolveMsg{
	Type:   ResponseLocked,
	From:   r.netName,
	Cid:    c,
	Locked: lstate,
}
return r.publishMsg(&m, to)
```

chain/consensus/hierarchical/subnet/resolver/resolver.go
```diff
@@ -281,20 +292,25 @@ func (r *Resolver) HandleIncomingResolveMsg(ctx context.Context, sub *pubsub.Sub
...
 	}
 	return pubsub.ValidationReject, xerrors.Errorf("Resolve message type is not valid")
 
 }
 
 func (r *Resolver) processPush(ctx context.Context, rmsg *ResolveMsg) (pubsub.ValidationResult, error) {
 	// Check if we are already storing the CrossMsgs CID locally.
-	_, found, err := r.getLocal(ctx, rmsg.Cid)
+	out := &sca.CrossMsgs{}
+	found, err := r.getLocal(ctx, rmsg.Cid, out)
```
dnkolegov, line 313: I would recommend:
```
out := sca.CrossMsgs{}
found, err := r.getLocal(ctx, rmsg.Cid, &out)
```

chain/consensus/hierarchical/subnet/resolver/resolver.go
```diff
@@ -281,20 +292,25 @@ func (r *Resolver) HandleIncomingResolveMsg(ctx context.Context, sub *pubsub.Sub
...
-		return r.processPull(submgr, rmsg)
-	case Response:
-		return r.processResponse(ctx, rmsg)
+		return r.processPullMeta(submgr, rmsg)
+	case ResponseMeta:
+		return r.processResponseMeta(ctx, rmsg)
+	case PullLocked:
+		return r.processPullLocked(submgr, rmsg)
+	case ResponseLocked:
+		return r.processResponseLocked(ctx, rmsg)
```
dnkolegov, line 304: Should we add default section as guard?

chain/consensus/hierarchical/subnet/manager/atomic.go
```diff
@@ -0,0 +1,361 @@
...
+	}
+	out := make([]sca.AtomicExec, 0)
+	for _, v := range m {
+		out = append(out, v)
+	}
+	return out, nil
+}
+
+func getAtomicExec(ctx context.Context, api *API, c cid.Cid) (*sca.AtomicExec, bool, error) {
+	var st sca.SCAState
```
dnkolegov, line 170: Maybe place it before the following line 
```
if err := pcst.Get(ctx, scaAct.Head, &st); err != nil {
	return nil, false, err
}
```

api/api_hierarchical.go
```diff
@@ -5,6 +5,7 @@ import (
 
 	"github.com/filecoin-project/go-address"
 	"github.com/filecoin-project/go-state-types/abi"
+	"github.com/filecoin-project/lotus/chain/consensus/hierarchical/actors/sca"
 	"github.com/filecoin-project/lotus/chain/consensus/hierarchical/checkpoints/schema"
 	"github.com/filecoin-project/lotus/chain/types"
 	"github.com/ipfs/go-cid"
```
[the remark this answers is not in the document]
adlrocha: Yes, I am afraid this is an inconsistency that we have throughout the whole repo. I will fix it in our Eudico-specific packages now, and defer the rest of the base code for the future (to make merge conflicts while rebasing with Lotus easier). Do you have any automatic tool to detect these inconsistencies?

chain/consensus/actors/atomic-replace/replace.go
```diff
@@ -33,45 +50,22 @@ var _ atomic.LockableActor = ReplaceActor{}
 // has illustrative purposes for now. Consider improving it in
 // future iterations.
 type ReplaceState struct {
-	Owners *atomic.LockedState
+	Owners    *atomic.LockedState
+	LockedMap cid.Cid //HAMT[cid]LockedState
```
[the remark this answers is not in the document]
adlrocha: We use it throughout all spec actors to specify that behind that CID there's a `HAMT[cid]LockedState`.

chain/consensus/hierarchical/subnet/resolver/resolver.go
adlrocha: Good catch, thanks.

chain/consensus/hierarchical/subnet/resolver/resolver.go
adlrocha: Fixed, but out of a curiosity, is this for performance or aesthetic reasons?

chain/consensus/hierarchical/subnet/resolver/resolver.go
adlrocha: We implicitly have it. If no case is matched we return an error and a `ValidationReject`.

chain/consensus/hierarchical/subnet/resolver/resolver.go
adlrocha: Fixed, but same as above, aesthetics or performance? Thanks!

api/api_hierarchical.go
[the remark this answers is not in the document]
dnkolegov: No, I don't. Let's just follow the policy and fix when we can.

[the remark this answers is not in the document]
adlrocha: I am fixing it to be (so we follow the same approach): 
```
    native
    not filecoin/eudico
    filecoin/eudico

```

chain/consensus/hierarchical/subnet/resolver/resolver.go
dnkolegov: This is to improve code readability.
The citation from Bill Kennedy's Go Playbook:
```
When constructing a variable, never use pointer semantics.
In other words, don’t construct to a pointer.
Construction tells you nothing about what the escape analysis algorithms will do.
It does not matter if the API you will be using wants a pointer to these values. Keep the data semantics clean.
The ampersand operator is an important readability marker for how a value is being shared.
```

This line of code tells you the value has the potential to be constructed on the heap since it is being shared up the call stack: `return &m`

This return tells you nothing without reading more code to understand what the `m` variable is: 
`return m`

Does it make sense?

dnkolegov (approved):

PR #144, opened 2022-03-16
Native support for hierarchical consensus in lotus and filcns

adlrocha:
- [x] Forked genesis package to be avoid import cycles and support the creating a genesis that initializes the SCA.
- [x] Support for cross-messages in Filcns and block validation.
- [x] Support for hierarchical consensus actor registry and cross-message validation in Filcns `TipSetExecutor`.
- [x] Lotus daemon support for subnets.
- [x] Include checkpoint buffer to support slow conensus algorithms in parents fixing a "nasty" bug in checkpoint commitment.

_This PR includes cross-messages for Filecoin consensus, which means that is no longer possible to connect an Eudico node with filecoin expected consensus to mainnet (without some code editing here and there). If you are looking to revert all the changes that break this compatibility you should blame this PR (mainly, the issue is in the inclusion of `CrossMessages` and the use of an extended `MsgMeta` that includes cross message cids)._


adlrocha:
@dnkolegov this is ready for review. Do not worry about the code in `chain/consensus/genesis/genesis.go` this is just a convenient copy-paste of `chain/gen/genesis` to prevent import cycles, so no need to review.

Issue #151, opened 2022-03-18
Add context to Consensus constructors


dnkolegov:
### Improvement Suggestion

If we have any component inside consensus then we will not be able to gracefully shutdown it.
We should add `context.Context` to the Consensus constructor to be able to do that.

PR #152, opened 2022-03-21
Add ctx into consensus constructors

dnkolegov:
This PS fixes [issue 151](https://github.com/filecoin-project/eudico/issues/151).


adlrocha (approved):
LGTM. Before merging, can you check if `go test ./...` passes in `chain/consensus/` as a sanity-check? Thanks!

PR #163, opened 2022-04-01
add tendermint to Garden

dnkolegov:
Add scripts to run Tendermint in Eudico Garden


adlrocha (approved):
LGTM. We have lots of scripts in the eudicogarde folder now. In the next PR it may be a good idea to think about organizing them in folders :) 

Let's test this early next week to see if we can have a cluster with Tendermint easily.

PR #168, opened 2022-04-07
Fix issue 162

dnkolegov:
This PR fixes https://github.com/filecoin-project/eudico/issues/162


chain/consensus/hierarchical/subnet/manager/manager.go
```diff
@@ -611,6 +612,29 @@ func (s *SubnetMgr) LeaveSubnet(
 	return smsg.Cid(), nil
 }
 
+func (s *SubnetMgr) ListSubnets(ctx context.Context) ([]sca.Subnet, error) {
```
adlrocha, line 615: This only lists the subnets for the rootnet. I would suggest adding a `subnetID` argument to the function so to choose the subnet from which to list the child subnets. Actually, strictly speaking this function is `ListChildSubnets` and not `ListSubnets` because it doesn't list all subnets in the hierarchy.

This was the case also for the command from which you took the code, my bad.

adlrocha (changes requested):
LGTM, but let's add a `subnetID` argument before merging for correctness. Thanks!

Issue #169, opened 2022-04-07
Add `ConsensusType` info to list-subnets command


adlrocha:
We currently don't include `ConsensusType` of the subnet in `./eudico subnet list-subnets`. It can be easily fixed and it would improve the UX.

PR #180, opened 2022-04-11
Fix issue 169

dnkolegov:
This PR fixes issue #169

### DIFF
--- a/cmd/eudico/subnet.go
+++ b/cmd/eudico/subnet.go
@@ -73,9 +73,12 @@ var listSubnetsCmd = &cli.Command{
 			if sh.Status != 0 {
 				status = "Inactive"
 			}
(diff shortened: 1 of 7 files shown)


adlrocha (changes requested):
Overall looks good. The only thing is that we shouldn't require the `SCAState` to have a `Consensus` field that is only used for this command. Let's fix it and ping me if you need an additional review. Thanks :)

cmd/eudico/subnet.go
```diff
@@ -325,7 +328,11 @@ var mineCmd = &cli.Command{
 		if err != nil {
 			return err
 		}
-		fmt.Fprintf(cctx.App.Writer, "Successfully started/stopped mining in subnet: %s\n", subnet)
+		if cctx.Bool("stop") {
```
adlrocha, line 331: Nice one!

PR #185, opened 2022-04-14
Mir intro

dnkolegov:
This PR adds MirBFT into Eudico


chain/consensus/hierarchical/actors/subnet/genesis.go
```diff
@@ -121,10 +122,19 @@ func WriteGenesis(
 		if err != nil {
 			return err
 		}
-		b, err = makeTendermintGenesisBlock(context.TODO(), bs, *template, checkPeriod)
+		b, err = makeTendermintGenesisBlock(ctx, bs, *template, checkPeriod)
 		if err != nil {
 			return xerrors.Errorf("error making genesis tendermint block: %w", err)
 		}
+	case hierarchical.MirBFT:
```
adlrocha, line 129: Let's make it an issue, but I am starting to feel that genesis generation is common for all consensus and we can probably extract it to a common function and simplify this code.

adlrocha: Thinking it through, I don't think it'll be a lot of work, in case you want to include this already in this PR.

chain/consensus/mirfbt/mine.go
```diff
@@ -0,0 +1,172 @@
...
+		msgBytes, err := msg.Serialize()
+		if err != nil {
+			log.Error("unable to serialize message:", err)
+			continue
+		}
+
+		reqNo := mirbft.ReqNo(msg.Message.Nonce)
+		tx := common.NewSignedMessageBytes(msgBytes, nil)
+
+		err = bft.mir.Node.SubmitRequest(ctx, mirbft.ClientID(0), reqNo, tx, nil)
```
adlrocha, line 105: Is there any relationship between `mirbft.ClientID` and the signing identity? (just a sanity-check, as I don't know how the mir-bft code works).

chain/consensus/mirfbt/mine.go
```diff
@@ -0,0 +1,172 @@
...
+
+	for _, msg := range bt.CrossMessages {
+		msgBytes, err := msg.Serialize()
+		if err != nil {
+			log.Error("unable to serialize cross-message:", err)
+			continue
+		}
+
+		tx := common.NewCrossMessageBytes(msgBytes, nil)
+		reqNo := mirbft.ReqNo(msg.Nonce)
```
adlrocha, line 122: Here several cross-messages may have the same nonce (as we discussed for Tendermint). This may be an issue for MirBFT (from our previous discussions). Just to have it in mind.

chain/consensus/mirfbt/mine.go
```diff
@@ -0,0 +1,172 @@
...
+		err = bft.mir.Node.SubmitRequest(ctx, mirbft.ClientID(0), reqNo, tx, nil)
+		if err != nil {
+			log.Error("unable to submit a message to MirBFT:", err)
+			continue
+		}
+		log.Debug("successfully sent a message to MirBFT")
+
+	}
+
+	// TODO: temporal delay to be able run tests.
```
adlrocha, line 133: Wait, why is this? This is a consequence of just using a single node? I don't get "not in sync with the subnet" part.

chain/consensus/mirfbt/node.go
```diff
@@ -0,0 +1,127 @@
...
+	"golang.org/x/xerrors"
+)
+
+const (
+	nodeBasePort = 10000
+)
+
+type mir struct {
+	Node  *mirbft.Node
+	OwnID t.NodeID
```
adlrocha, line 28: Suggestion: make this a peer.ID

chain/consensus/mirfbt/node.go
```diff
@@ -0,0 +1,127 @@
...
+	Wal   *simplewal.WAL
+	Net   *grpctransport.GrpcTransport
```
adlrocha, line 30: Suggestion: Figure out how to make this a general interface where we can hook libp2p instead of making it grpc specific.

chain/consensus/mirfbt/node.go
```diff
@@ -0,0 +1,127 @@
...
+			close(stopC)
+		case <-nodeCtx.Done():
+			n.stop()
+		}
+
+		log.Info("MirBFT node was stopped")
+	}()
+
+	go func() {
+		// TODO: add a bug into the MirBFT repo: use context for signalling
```
adlrocha, line 114: +1

chain/consensus/mirfbt/utils.go
```diff
@@ -0,0 +1,34 @@
...
+import (
+	"golang.org/x/xerrors"
+
+	"github.com/filecoin-project/lotus/chain/consensus/common"
+	"github.com/filecoin-project/lotus/chain/types"
+)
+
+func parseTx(tx []byte) (interface{}, error) {
+	ln := len(tx)
+	if ln <= 2 {
```
adlrocha, line 12: Why <= 2?

adlrocha (approved):
Awesome job, @dnkolegov. I added a bunch of comments to understand better the code, as well as a few suggestions for future work (especially, let's try to figure out a good way to avoid a lot of the duplicate code we have in several consensus implementations). But in any case, this is in a good state so feel free to merge :pray:.

chain/consensus/mirfbt/utils.go
dnkolegov: Actually < 2. Because the last byte is message type. It would be great to improve this validation.

chain/consensus/hierarchical/actors/subnet/genesis.go
dnkolegov: The code is different for delegated. I created an issue https://github.com/filecoin-project/eudico/issues/187

chain/consensus/mirfbt/mine.go
dnkolegov: Signing identity is a node identity, as far as I know. In future, client ID will be equal to Eudico peer ID

chain/consensus/mirfbt/mine.go
dnkolegov: added into TODO.

chain/consensus/mirfbt/mine.go
dnkolegov: Blocks in the root net (with MirBFT) are mined very fast. When we use WaitSubnetActorBalance with limit = 100, these 100 blocks are mined faster when events are processed in the subnet.

But I fixed that

matejpavlovic (commented):
Thanks for all the work!
I only really reviewed the parts related to Mir that I understand, so I left most of the files untouched.
I have a few general questions / comments:

- What is ideal consensus?
- It would be nicer (and easier for me to review) if there was a separate PR just for the Mir-related code.
- Could you add a few comments in the code, saying (in just a few words) what the code is doing? That would also make it easier for me to read.

Regarding the way of delivering blocks from the App to Eudico, I realized that there might be a problem with the approach and we might need a bit more synchronization, but if we assume no restarts, it should be fine for now. I'll create another separate issue for that.

chain/consensus/mir/mine.go
```diff
@@ -0,0 +1,190 @@
...
+			continue
+		}
+
+		select {
+		case merr := <-mirErrors:
+			return xerrors.Errorf("Mir node error: %w", merr)
+		case <-ctx.Done():
+			log.Debug("Mir miner %s: context closed")
+			return nil
+		case mirBlock := <-mirHead:
```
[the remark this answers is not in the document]
dnkolegov: improved

chain/consensus/mir/mine.go
```diff
@@ -0,0 +1,190 @@
...
+	"time"
+
+	"golang.org/x/xerrors"
+
+	"github.com/filecoin-project/go-address"
+	lapi "github.com/filecoin-project/lotus/api"
+	"github.com/filecoin-project/lotus/api/v1api"
+	"github.com/filecoin-project/lotus/chain/consensus/common"
+	"github.com/filecoin-project/lotus/chain/types"
+	mir "github.com/filecoin-project/mir/pkg/types"
```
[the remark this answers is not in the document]
dnkolegov: fixed

chain/consensus/mir/mir_agent.go
```diff
@@ -0,0 +1,165 @@
...
+	// Instantiate the ISS protocol module with default configuration.
+	issConfig := iss.DefaultConfig(nodeIds)
+	issProtocol, err := iss.New(ownID, issConfig, newMirLogger(agentLog))
+	if err != nil {
+		return nil, xerrors.Errorf("could not instantiate ISS protocol module: %w", err)
+	}
+
+	app := NewApplication(reqStore)
+
+	// TODO: write a wrapper on Eudico logger to use it in Mir.
```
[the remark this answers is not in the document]
dnkolegov: ok

chain/consensus/mir/mir_agent.go
```diff
@@ -0,0 +1,165 @@
...
+func newMirLogger(logger *logging.ZapEventLogger) *mirLogger {
+	return &mirLogger{
+		logger: logger,
+	}
+}
+
+func (m *mirLogger) Log(level mirLogging.LogLevel, text string, args ...interface{}) {
+	switch level {
+	case mirLogging.LevelError:
+		m.logger.Errorf(text, args)
```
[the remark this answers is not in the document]
dnkolegov: fixed

chain/consensus/mir/app.go
```diff
@@ -0,0 +1,79 @@
...
+	"github.com/filecoin-project/mir/pkg/pb/requestpb"
+)
+
+type Tx []byte
+
+type Application struct {
+	mu sync.Mutex
+
+	reqStore modules.RequestStore
+	height   int64
```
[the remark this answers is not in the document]
dnkolegov: removed

chain/consensus/mir/app.go
```diff
@@ -0,0 +1,79 @@
...
+	"sync"
+
+	"github.com/filecoin-project/mir/pkg/modules"
+	"github.com/filecoin-project/mir/pkg/pb/requestpb"
+)
+
+type Tx []byte
+
+type Application struct {
+	mu sync.Mutex
```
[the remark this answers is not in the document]
dnkolegov: It depends on the request store and how many clients share the application. If all structures are unique then we don't need this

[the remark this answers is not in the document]
matejpavlovic: The App abstraction is meant to only be used as a module for a single instance of Mir, so it looks like we don't need to synchronize access to the `Apply` method. This is not to say that the object implementing the `App` interface cannot interact with the rest of the system outside of Mir - it has to, in fact, otherwise it wouldn't be very useful. It's just that the `Apply` function will never be called concurrently. A mutex might still make sense if the `Apply` method accesses some data structures that are also accessed through other methods not called by Mir, but that doesn't seem to be the case now.

chain/consensus/mir/mine.go
```diff
@@ -0,0 +1,190 @@
...
+				Timestamp:        uint64(time.Now().Unix()),
+				WinningPoStProof: nil,
+				Messages:         msgs,
+				CrossMessages:    crossMsgs,
+			})
+			if err != nil {
+				log.Errorw("creating block failed", "error", err)
+				continue
+			}
+			if bh == nil {
```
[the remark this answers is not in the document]
matejpavlovic: Hm but doesn't that mean that we lose one batch of transactions that we've just received from Mir? What happens to the transactions jus read from the channel in the select statement?

dnkolegov:
Thank you for your review.

> * What is ideal consensus?

It is a primitive for testing: it orders input messages by adding them into a block in the same order they were received, it works only for one tested instance.

> * It would be nicer (and easier for me to review) if there was a separate PR just for the Mir-related code.

Sorry for that. But adding Mir required fixing other stuff too. See Alfonso's suggestions and comments.


> * Could you add a few comments in the code, saying (in just a few words) what the code is doing? That would also make it easier for me to read.

Sure. What part of the code you didn't understand?

chain/consensus/mir/app.go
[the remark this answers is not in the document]
dnkolegov: ok. removed.

chain/consensus/mir/mine.go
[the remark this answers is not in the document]
dnkolegov: Yes, correct.
We received transactions from App, tried to create a block, and failed, for example, because one tx was invalid or we were unable to sign the block.
In the current design, these all transactions will not be removed from mempool and will be sent later.

matejpavlovic:
> Sure. What part of the code you didn't understand?

Mostly `mine.go`, `mir_agent.go` and `utils.go` have almost no comments. It's not that I didn't understand them at all, but I think it would make the code easier to read if a few comments were saying in English what the lines of code are doing.
The other files, I assume, are mostly boilerplate required by Eudico that we don't care about too much yet.

But otherwise all is fine with me, if it works (and when @adlrocha submits his review) we can merge and go on implementing support for multiple nodes.

chain/consensus/hierarchical/actors/subnet/genesis.go
adlrocha: LGTM. Closing the issue.

chain/consensus/ideal/ideal.go
```diff
@@ -0,0 +1,241 @@
+// Package ideal implements Ideal consensus for testing purposes only.
```
adlrocha, line 1: Isn't ideal quite redundant with delegated? What about removing delegated and keeping ideal as the "debug" consensus (or vice-versa?). 

If we go this way I would maybe call it "debug" instead of "ideal" consensus. We would have to add cmds to be able to run it as part of the daemon, but I think this may be worth the effort.

chain/consensus/mir/mine.go
```diff
@@ -0,0 +1,185 @@
...
+// 1. Retrieve messages and cross-messages from mempool.
+// 2. Send these messages to the Mir node.
+// 3. Receive ordered messages from the Mir node and push them into the next Filecoin block.
+// 4. Submit this block.
+func Mine(ctx context.Context, miner address.Address, api v1api.FullNode) error {
+	log.Info("Mir miner started")
+	defer log.Info("Mir miner stopped")
+
+	// TODO: Suppose we want to use Mir in Root and in a subnet at the same time.
+	// Do we need two Mir agents for that?
```
adlrocha, line 28: Probably? I would have to look deeper on how Mir is implemented, but probably that is the cleanest approach. Probably @matejpavlovic can clarify this one.

chain/consensus/mir/mine.go
```diff
@@ -0,0 +1,185 @@
...
+		if err != nil {
+			log.Errorw("failed to get the head of chain", "error", err)
+			continue
+		}
+
+		select {
+		case <-ctx.Done():
+			log.Debug("Mir miner: context closed")
+			return nil
+		case merr := <-mirErrors:
```
adlrocha, line 65: I like this approach to kill the mining process. Do you think it make sense to propagate it to all consensus algorithms?

Context: As we discussed sync, our consensus algorithms are architected a bit differently to how `filcns` is, and for subnets we don't run miners in a separate process. This would allow us to terminate the mining process if there is an error instead of looping with errors forever (my only concern with this would be if there are recoverable errors that don't lead to an infinite looping error in mining). I leave the final decision to you :)

chain/consensus/mir/mine.go
```diff
@@ -0,0 +1,185 @@
...
+
+	netName, nerr := api.StateNetworkName(ctx)
+	if nerr != nil {
+		return nerr
+	}
+	subnetID := address.SubnetID(netName)
+
+	log.Infof("Mir miner params: network name - %s, subnet ID - %s", netName, subnetID)
+
+	submit := time.NewTicker(SubmitInterval)
```
adlrocha, line 51: Will this ticker still be needed once Mir is fully implemented? If this is the case, we may want to add a TODO to figure out the optimal ticker to maximize performance.

chain/consensus/mir/mine.go
```diff
@@ -0,0 +1,185 @@
...
+				msgBytes, err := msg.Serialize()
+				if err != nil {
+					log.Error("unable to serialize message:", err)
+					continue
+				}
+
+				reqNo := mirTypes.ReqNo(msg.Message.Nonce)
+				tx := common.NewSignedMessageBytes(msgBytes, nil)
+
+				// TODO: define what client ID is in Eudico case.
```
adlrocha, line 130: Yes. I'll shy away of requiring different IDs. The right interface in MIR may do the trick.

chain/consensus/mir/mir_agent.go
```diff
@@ -0,0 +1,163 @@
...
+type MirAgent struct {
+	Node     *mir.Node
+	Wal      *simplewal.WAL
+	Net      *grpctransport.GrpcTransport
+	App      *Application
+	stopChan chan struct{}
+}
+
+func NewMirAgent(id string, n int) (*MirAgent, error) {
+	// TODO: Are client ID and node ID the same in this case?
```
adlrocha, line 62: @matejpavlovic? This is more a conceptual question, right?

cmd/eudico/mir.go
```diff
@@ -0,0 +1,115 @@
...
+	Name:  "mir",
+	Usage: "Mir consensus",
+	Subcommands: []*cli.Command{
+		mirbftGenesisCmd,
+		mirbftMinerCmd,
+
+		daemonCmd(node.Options(
+			node.Override(new(consensus.Consensus), NewRootMirConsensus),
+			node.Override(new(store.WeightFunc), mir.Weight),
+			node.Unset(new(*slashfilter.SlashFilter)),
```
adlrocha, line 46: out of a curiosity, why are we unsetting the slash filter here? (I agree that is not needed, but we haven't unset it in any other consensus) :)

adlrocha (approved):
LGTM. Left a few comments but I don't think they are blocking (actually they may be quite redundant with @matejpavlovic's review, I haven't looked in depth to his review). Feel free to disregard them. 

One thing maybe worth discussing for "short-term" future work:
- Why not removing delegated consensus for the ideal (or debug) consensus? They are quite similar, and what we are going for with both is to have a dumb consensus algorithm we can use for debugging and testing purposes. This would require adding support for "ideal" in subnets and to add support to run it as a daemon in rootnet (but I don't think this is hard and we would clean a lot of redundant code). Let's add an issue if we are all happy with this.

chain/consensus/ideal/ideal.go
dnkolegov: The idea of Ideal consensus is to have the simplest and most independent consensus protocol. The delegated consensus is more complex, at least its instantiation is not trivial, it can be spawned on several nodes. I would say that the delegation consensus is the next level: we start with the simplest reference model (ideal/debug consensus) and move forward with tests to be sure we test the target code (not consensus code) with ideal dependencies.

chain/consensus/mir/mine.go
dnkolegov: Created an issue https://github.com/filecoin-project/eudico/issues/188

cmd/eudico/mir.go
dnkolegov: It was unset in PoW, that was the reference impl for all next consensus. Should I set SlashFilter in all of them?

chain/consensus/mir/mine.go
matejpavlovic: Yes I think that is the way to do it. Unless we want to somehow explicitly share the consensus mechanism for both the root and the subnet (in which case it's questionable why we even have a subnet), there should be two distinct mir agents.

chain/consensus/ideal/ideal.go
dnkolegov: That's all correct. I called its ideal following the Real/Ideal paradigm. No problems to rename it

chain/consensus/mir/mine.go
dnkolegov: @matejpavlovic will be happy to answer

cmd/eudico/mir.go
dnkolegov: And in DelegCns there is no SlashFilter at all

adlrocha: Not critical, it was just out of a curiosity.

chain/consensus/mir/mine.go
matejpavlovic: We should not need that ticker later - as soon as the mempool does not need to be polled for new messags. One of the points of the ordering layer architecture that I'm currently drafting is to modify the mempool to be able to create notifications about newly received messages (e.g. by writing them to a channel). When we have that, this timer can go away.

That being said, the underlying protocol ordering the messages will definitely require some form of batching, which, depending on the protocol, might or might not involve some periodic timeouts. But that will be the job of the responsibility of the protocol implementation.

chain/consensus/mir/mine.go
matejpavlovic: If I understood correctly from previous discussions, each message has a "nonce", which is a monotonically increasing number (I'd prefer calling it a "message number", "request number", or "sequence number", but as long as we all know what we're talking about, I don't care that much). This nonce uniquely identifies a message among all the messages coming, in a way, from the same "source". This source is something that produces a sequence of messages and assigns them the nones. In other words, the source defines a set of messages among which nonces are unique. The identifier of this source is what we need to use as "client ID". What is this identifier in Eudico terms? Actor address? Wallet address?

chain/consensus/mir/mir_agent.go
matejpavlovic: Copy-pasting from my answer to the previous comment:
> If I understood correctly from previous discussions, each message has a "nonce", which is a monotonically increasing number (I'd prefer calling it a "message number", "request number", or "sequence number", but as long as we all know what we're talking about, I don't care that much). This nonce uniquely identifies a message among all the messages coming, in a way, from the same "source". This source is something that produces a sequence of messages and assigns them the nones. In other words, the source defines a set of messages among which nonces are unique. The identifier of this source is what we need to use as "client ID".

As for node ID, it can be anything that uniquely identifies a node that participates in the consensus protocol. The Eudico node's ID seems like a natural choice to me here.

dnkolegov: We discussed with @adlrocha : subnetName + wallet address can be used as IDs.

chain/consensus/mir/mine.go
dnkolegov: We discussed with @adlrocha : subnetName + wallet address can be used as IDs.

chain/consensus/mir/mir_agent.go
matejpavlovic: Sounds reasonable.

chain/consensus/dummy/dummy.go
```diff
@@ -0,0 +1,241 @@
...
+	sm *stmgr.StateManager,
+	submgr subnet.SubnetMgr,
+	b beacon.Schedule,
+	r *resolver.Resolver,
+	v ffiwrapper.Verifier,
+	g chain.Genesis,
+	netName dtypes.NetworkName,
+) (consensus.Consensus, error) {
+	subnetID := address.SubnetID(netName)
+	log.Infof("New Mir consensus for %s subnet", subnetID)
```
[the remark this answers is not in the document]
dnkolegov: thanks.

PR #189, opened 2022-04-27
Add Mir persistent nodes

dnkolegov:
This PR adds Mir persistent nodes similar to [Tendermint persistent peers](https://docs.tendermint.com/master/introduction/quick-start.html) to be able to run multiple Eudico nodes on remote hosts.


adlrocha (approved):
LGTM. This, however, means that we have the same set of validators for all subnets, right? But for demo purposes that's OK, let's improve it once we figure out the right permissioning approach as discussed sync.

matejpavlovic (approved):
Great idea to use the environment for ad-hoc configuration. Yes it has its limits as Alfonso pointed out, but for now it's simple and efficient, and later (hopefully quite soon) we will have a proper concept of (re)configuration that will deal with this properly.

adlrocha (changes requested):
Quick note, this PR has broken `JoinSubnets` unit tests and is making the linter unhappy. Please remember to fix it before merging :) Thanks!

api/api_hierarchical.go
```diff
@@ -13,28 +13,29 @@ import (
...
+	SubnetChainNotify(context.Context, address.SubnetID) (<-chan []*HeadChange, error)                                                                                                                                      // perm:read
+	SubnetChainHead(context.Context, address.SubnetID) (*types.TipSet, error)                                                                                                                                               // perm:read
+	SubnetStateGetActor(ctx context.Context, id address.SubnetID, addr address.Address, tsk types.TipSetKey) (*types.Actor, error)                                                                                          // perm:read
+	SubnetStateWaitMsg(ctx context.Context, id address.SubnetID, cid cid.Cid, confidence uint64, limit abi.ChainEpoch, allowReplaced bool) (*MsgLookup, error)                                                              // perm:read
+	SubnetGetActorStateValidators(ctx context.Context, id address.SubnetID) (string, error)                                                                                                                                 // perm:read
```
adlrocha, line 40: What about calling it `SubnetGetValidators`?

dnkolegov: fixed

api/api_hierarchical.go
```diff
@@ -13,28 +13,29 @@ import (
...
-	SubnetChainNotify(context.Context, address.SubnetID) (<-chan []*HeadChange, error)                                                                                                                                          // perm:read
-	SubnetChainHead(context.Context, address.SubnetID) (*types.TipSet, error)                                                                                                                                                   // perm:read
-	SubnetStateGetActor(ctx context.Context, id address.SubnetID, addr address.Address, tsk types.TipSetKey) (*types.Actor, error)                                                                                              // perm:read
-	SubnetStateWaitMsg(ctx context.Context, id address.SubnetID, cid cid.Cid, confidence uint64, limit abi.ChainEpoch, allowReplaced bool) (*MsgLookup, error)                                                                  // perm:read
+	AddSubnet(ctx context.Context, wallet address.Address, parent address.SubnetID, name string, cns uint64, minerStake abi.TokenAmount, checkperiod abi.ChainEpoch, dm address.Address, n uint64) (address.Address, error) // perm:write
```
adlrocha, line 16: Suggestion (not critical to merge but worth at least considering for the future): what about aggregating all the optional arguments which are consensus-specific in a struct? We could have something like: 
```go
type ConsOpts struct {
  delegMiner
  dm address.Address
  etc.
}
``` 
It may make the function signature a bit cleaner.

chain/consensus/hierarchical/actors/subnet/iface.go
```diff
@@ -1,16 +1,15 @@
...
-
+	"github.com/filecoin-project/go-state-types/abi"
 	"github.com/filecoin-project/lotus/chain/consensus/hierarchical/actors/sca"
 	"github.com/filecoin-project/specs-actors/v7/actors/runtime"
 )
 
-// SubnetIface defines the minimum interface that needs to be implemented by subnet
-// actors
+// SubnetIface defines the minimum interface that needs to be implemented by subnet actors.
+// TODO: @alfonso do we need this?
```
adlrocha, line 10: The idea was for this interface to be the one that users would need to implement to deploy their own subnet actors. This doesn't make much sense in the current scenario where there's no support for user-defined actors, but it'll change with FVM. 

If you feel that constantly fixing this interface with every change is making you unproductive feel free to remove it, but I'd rather keep it there for completeness until we have the FVM actors ready and we remove all this code.

chain/consensus/hierarchical/actors/subnet/subnet_actor.go
```diff
@@ -64,12 +65,13 @@ func (a SubnetActor) State() cbor.Er {
...
-	MinMinerStake abi.TokenAmount            // MinStake to give miner rights
-	DelegMiner    address.Address            // Miner in delegated consensus
-	CheckPeriod   abi.ChainEpoch             // Checkpointing period.
+	NetworkName      string                     // Name of the current network.
+	Name             string                     // Name for the subnet
+	Consensus        hierarchical.ConsensusType // Consensus for subnet.
+	MinMinerStake    abi.TokenAmount            // MinStake to give miner rights
+	DelegMiner       address.Address            // Miner in delegated consensus
+	CheckPeriod      abi.ChainEpoch             // Checkpointing period.
+	ValidatorsNumber uint64                     // Number of validators.
```
adlrocha, line 74: Would it be correct to say that this is not `ValidatorsNumber` but `MinValidators`, i.e. the minimum number of validators for the subnet to be able to start mining? (I know that this is a constant right now, but it'll change once we support reconfiguration, and that way we can leverage it for other consensus algorithms, and it allows us to even add this check for other consensus algorithms (like PoW)).

Happy to discuss sync if needed.

chain/consensus/hierarchical/actors/subnet/subnet_actor.go
```diff
@@ -143,6 +145,18 @@ func (a SubnetActor) Join(rt runtime.Runtime, _ *abi.EmptyValue) *abi.EmptyValue
 		}
 	}
 
+	rt.StateTransaction(&st, func() {
+		// Mutate state
+		if st.Consensus == hierarchical.Mir {
```
adlrocha, line 150: Why not making this functionality general and not Mir-specific? We can set `MinValidators` to be 0 by default, but having this for every consensus algorithm would enable interesting use cases (like for instance requiring a minimum number of PoW miners before starting a network to prevent the first joiner from having an unfair advantage).

chain/consensus/hierarchical/actors/subnet/subnet_state.go
```diff
@@ -80,6 +79,10 @@ type SubnetState struct {
 	Checkpoints cid.Cid // HAMT[epoch]Checkpoint
 	// WindowChecks
 	WindowChecks cid.Cid // HAMT[cid]CheckVotes
+	// Validators contains information about BFT validators
+	Validators []Validator
+	// ValidatorsNumber initial number of validators to start a subnet with BFT-type consensus
+	ValidatorsNumber uint64
```
adlrocha, line 85: MinValidators? (the current number of validators can be checked through `len(Validators)` (see comment above)

chain/consensus/hierarchical/actors/subnet/validator.go
```diff
@@ -0,0 +1,65 @@
...
+
+	"golang.org/x/xerrors"
+
+	t "github.com/filecoin-project/mir/pkg/types"
+)
+
+// Validator encapsulated string to be compatible with CBOR implementation.
+type Validator struct {
+	ID      string
+	Address string
```
adlrocha, line 15: I would expect this to be a multiaddr almost always, but I am fine with this general approach.

chain/consensus/hierarchical/subnet/manager/manager.go
```diff
@@ -350,12 +351,29 @@ func BuildSubnetMgr(mctx helpers.MetricsCtx, lc fx.Lifecycle, s *SubnetMgr) {
 	})
 }
 
+func (s *SubnetMgr) GetActorState(ctx context.Context, id address.SubnetID, actor address.Address) (*subnet.SubnetState, error) {
```
adlrocha, line 354: As discussed sync, it is better to use `api := GetSubnetAPI(...)` and from there `GetActorState`, than exposing this function in all APIs (from the root to ever subnet) publically.

chain/consensus/mir/mine.go
```diff
@@ -2,52 +2,73 @@ package mir
...
 // Mine mines Filecoin blocks.
-
+//
 // Mine implements the following algorithm:
 // 1. Retrieve messages and cross-messages from mempool.
 // 2. Send these messages to the Mir node.
 // 3. Receive ordered messages from the Mir node and push them into the next Filecoin block.
 // 4. Submit this block.
+//
+// There are two ways how mining with Mir consensus can be launching:
```
adlrocha, line 28:
```suggestion
// There are two ways mining can be launched for Mir consensus:
```

node/impl/hierarchical/subnet.go
```diff
@@ -28,13 +28,14 @@ func (a *HierarchicalAPI) AddSubnet(
 	parent address.SubnetID, name string,
 	consensus uint64, minerStake abi.TokenAmount,
 	checkPeriod abi.ChainEpoch,
-	delegminer address.Address) (address.Address, error) {
-	return a.Sub.AddSubnet(ctx, wallet, parent, name, consensus, minerStake, checkPeriod, delegminer)
+	dm address.Address,
```
adlrocha, line 31: Same as above, we should probably make a struct of opts to make the function signature more readable and dev-friendly.

cmd/eudico/subnet.go
```diff
@@ -271,7 +293,9 @@ var syncCmd = &cli.Command{
 		if err != nil {
 			return err
 		}
-		fmt.Fprintf(cctx.App.Writer, "Successfully started/stopped syncing with subnet %s \n", subnet)
+		if _, err := fmt.Fprintf(cctx.App.Writer, "Successfully started/stopped syncing with subnet %s \n", subnet); err != nil {
```
adlrocha, line 296: I don't think this is right, we don't care about the Fprintf error, we can forget about it, right? (see also changes below)

adlrocha (approved):
LGTM. No major blockers.The thing I would like to at least address before merging is the `Fprintf` error change you included for cli commands. I don't think we should return the Fprintf err.

For the rest, feel free to address the comments you agree with, and defer the things that you don't see as critical (although I think they should be quite straightforward).

Thanks! Awesome job.

chain/consensus/hierarchical/actors/subnet/iface.go
dnkolegov: ok

chain/consensus/hierarchical/actors/subnet/validator.go
dnkolegov: fixed

chain/consensus/hierarchical/actors/subnet/subnet_actor.go
dnkolegov: fixed

chain/consensus/hierarchical/actors/subnet/subnet_actor.go
dnkolegov: fixed

api/api_hierarchical.go
dnkolegov: fixed

adlrocha (approved):
LGTM. Feel free to merge

PR #193, opened 2022-05-11
Preparing for Mir demo

dnkolegov:
This PR adds several useful mechanisms for demo purposes (e.g., logging in a subnet), prepares new scripts, and fixes minor bugs


chain/consensus/platform/logging/logger.go
```diff
@@ -0,0 +1,51 @@
...
+		return logger
+	}
+	return logger
+}
+
+func WithLogger(ctx context.Context, logger *logging.ZapEventLogger) context.Context {
+	return context.WithValue(ctx, loggerKey, logger)
+}
+
+func NewFileLogger(level string, name string) (*logging.ZapEventLogger, error) {
```
adlrocha, line 26: Just to see if I've understood this: the reason why you use low-level `zap` insteado of using directly the ipfs-logger and setting `GOLOG_OUTPUT` is because you want to output mining logs into an independent file, right? (https://github.com/ipfs/go-log#golog_output)

chain/consensus/hierarchical/types.go
```diff
@@ -18,6 +18,11 @@ type ConsensusParams struct {
 	MinValidators uint64          // Min number of validators required to start a network.
 }
 
+type MiningParams struct {
```
adlrocha, line 21: Should we call this something like `MiningLogParams`, or do you expect to use these params for something else in the future?

chain/consensus/hierarchical/gen/gen.go
```diff
@@ -10,6 +10,7 @@ func main() {
 	if err := gen.WriteTupleEncodersToFile("./cbor_gen.go", "hierarchical",
 		hierarchical.ConsensusParams{},
 		hierarchical.Validator{},
+		hierarchical.MiningParams{},
```
adlrocha: Why do you need `MiningParams` to be CBOR-serializable? I don't think you are using it anywhere where this struct needs to implement `CBORMarshaler` (or maybe I've missed it). Is it because you are using it as an argument in `Mine` for the API?

adlrocha (changes requested):
LGTM. The only thing to check before merging is why are we implementing `CBORMarshaler` for `MiningParams` (i.e. why are we including it in `gen.go`), it is so it can be serialized as an argument for the API?

api/api_hierarchical.go
```diff
@@ -17,7 +17,7 @@ type HierarchicalCns interface {
 	AddSubnet(ctx context.Context, wallet address.Address, parent address.SubnetID, name string, cns uint64, minerStake abi.TokenAmount, checkPeriod abi.ChainEpoch, p *hierarchical.ConsensusParams) (address.Address, error) // perm:write
 	JoinSubnet(ctx context.Context, wallet address.Address, value abi.TokenAmount, id address.SubnetID, validatorAddr string) (cid.Cid, error)                                                                                 // perm:write
 	SyncSubnet(ctx context.Context, id address.SubnetID, stop bool) error                                                                                                                                                      // perm:write
-	MineSubnet(ctx context.Context, wallet address.Address, id address.SubnetID, stop bool) error                                                                                                                              // perm:read
+	MineSubnet(ctx context.Context, wallet address.Address, id address.SubnetID, stop bool, p *hierarchical.MiningParams) error                                                                                                // perm:read
```
adlrocha, line 20: Now that I see the permissions of this function, maybe we should change them to `perm:write`. What do you think?

chain/consensus/hierarchical/types.go
dnkolegov: I thought we will add other parameters in the future. So it's not only for log parameters

chain/consensus/hierarchical/gen/gen.go
dnkolegov: Yes, I thought we have to serialized all API input

api/api_hierarchical.go
dnkolegov: It depends on what `perm:write` means. We don't change the system state. Right? We outputted log and before that... but that was another place.

chain/consensus/hierarchical/gen/gen.go
adlrocha: I don't think you need this. Try removing this, re-generate the cbor-gen and see if it works. IIRC there's no need for this if you are only going to use it as an API argument.

chain/consensus/hierarchical/types.go
adlrocha: Gotcha, the this is fine by me.

api/api_hierarchical.go
adlrocha: Yes, my question is not with your change, but that when reading it I just realized that maybe that mine should have write instead of read permissions. Is not critical.

chain/consensus/platform/logging/logger.go
dnkolegov: Yes, correct. `GOLOG_OUTPUT` doesn't allow to specify the concrete file for the target log stream.

adlrocha (approved):
LGTM.

PR #199, opened 2022-05-18
Reference Rust implementation for subnet actor

adlrocha:
This PR includes the reference Rust implementation of the subnet actor to target the FVM. This actor is user-defined, and users are allowed to fork it and implement it as they wish. They only requirement we'll impose is that it implements the `SubnetActor` trait.
- [ ] Subnet actor functions
  - [x] Constructor
  - [ ] Join
  - [ ] Kill
  - [ ] Leave
  - [ ] SubmitCheckpoint 
- [ ] `SubnetActor` trait
- [ ] Unit tests

_NOTE: This implementation may differ slightly from the Go implementation and it'll require changes over Eudico and the way it interacts with the actor (but this will also be the case for the SCA and other hierarchical consensus-related assets)_


adlrocha:
Actor being deployed in the current repo as a standalone project for now to improve code management (once it is fully implemented the actor may make its way back this repo): https://github.com/adlrocha/hc-subnet-actor

PR #200, opened 2022-05-18
Update CI config to be able to run race detector

### LINKED ISSUE #198: Run tests with race flag to detect data races
### Improvement Suggestion

Run tests in CI with the `race` flag to detect data races.
Our first attempt to do that revealed we need a more sophisticated approach.
The basic idea is to run tests with the `race` flag for different `_test.go` files sequentially.

dnkolegov:
This PR fixes https://github.com/filecoin-project/eudico/issues/198

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -806,7 +806,6 @@ workflows:
       - test_subnet_races:
           requires:
             - lint-all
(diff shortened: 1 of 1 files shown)


adlrocha (approved):
LGTM

PR #202, opened 2022-05-24
Initial Eudico ordering interface

dnkolegov:
This PR is a work in progress implementation of this interface - https://hackmd.io/6v00B72USpav6rDUUhqxAw


chain/consensus/mir/app.go
```diff
@@ -23,6 +23,11 @@ func NewApplication(reqStore modules.RequestStore) *Application {
 func (app *Application) Apply(batch *requestpb.Batch) error {
 	var block []Tx
 
+	if len(batch.Requests) == 0 {
+		app.ChainNotify <- block
```
matejpavlovic: If the batch is empty, we might decide not to write the block to the channel at all. If we do need to write it (e.g. because Eudico needs the empty blocks as well), then let's completely skip this `if` block, as it looses its meaning.

chain/consensus/mir/manager.go
```diff
@@ -0,0 +1,299 @@
...
+func (m *mirLogger) Log(level mirLogging.LogLevel, text string, args ...interface{}) {
+	switch level {
+	case mirLogging.LevelError:
+		m.logger.Errorw(text, "error", args)
+	case mirLogging.LevelInfo:
+		m.logger.Infow(text, "info", args)
+	case mirLogging.LevelWarn:
+		m.logger.Warnw(text, "warn", args)
+	case mirLogging.LevelDebug:
+		m.logger.Debugw(text, "debug", args)
```
matejpavlovic, line 53: It looks like the logger has a Zap-like interface, where the first argument is the log message and the rest represents a list of key/value pairs. In such case, the string literal (`"error"`, `"info"`, etc...) seems superfluous and, in fact offsets the rest of the arguments so that keys and values are swapped, the first element of `args` effectively becoming the "value" for the `"error"` key. Is that the case? I might be missing something though...

chain/consensus/mir/manager.go
```diff
@@ -0,0 +1,299 @@
...
+		if subnetID == address.RootSubnet {
+			return nil, xerrors.New("can't be run in rootnet without validators")
+		}
+		validators, err = api.SubnetStateGetValidators(ctx, subnetID)
+		if err != nil {
+			return nil, xerrors.New("failed to get validators from state")
+		}
+	}
+	if len(validators) == 0 {
+		return nil, xerrors.New("empty validator set")
```
matejpavlovic, line 109: Use `fmt.Errorf` here? In other parts of the code that seems to be the convention.

chain/consensus/mir/manager.go
```diff
@@ -0,0 +1,299 @@
...
+	wal, err := NewWAL(mirID, "eudico-wal")
+	if err != nil {
+		return nil, err
+	}
+
+	net := grpctransport.NewGrpcTransport(nodeAddrs, t.NodeID(mirID), newMirLogger(managerLog))
+	if err := net.Start(); err != nil {
+		return nil, err
+	}
+	net.Connect(ctx)
```
matejpavlovic, line 135: It's probably worth a comment here that the GrpcTransport is a rather dummy one and this call blocks until **all** connections are established, which makes it, at this point, not fault-tolerant.

chain/consensus/mir/manager.go
```diff
@@ -0,0 +1,299 @@
...
+	"github.com/filecoin-project/mir/pkg/iss"
+	mirLogging "github.com/filecoin-project/mir/pkg/logging"
+	"github.com/filecoin-project/mir/pkg/modules"
+	"github.com/filecoin-project/mir/pkg/reqstore"
+	"github.com/filecoin-project/mir/pkg/simplewal"
+	t "github.com/filecoin-project/mir/pkg/types"
+)
+
+// managerLog is a Eudico logger used by a Mir node.
+var managerLog = logging.Logger("mir-manager")
```
matejpavlovic, line 30: I wonder whether this shouldn't be made a property of the Manager rather than a global package field... In case we need to run more than one instance of the manager at some point.

chain/consensus/mir/manager.go
```diff
@@ -0,0 +1,299 @@
...
+		Net:        net,
+		App:        app,
+	}
+
+	return &a, nil
+}
+
+// Start starts an agent.
+func (m *Manager) Start(ctx context.Context) chan error {
+	log.Info("Mir agent starting")
```
matejpavlovic: Manager instead of agent?

chain/consensus/mir/manager.go
```diff
@@ -0,0 +1,299 @@
...
+func NewManager(ctx context.Context, addr address.Address, api v1api.FullNode) (*Manager, error) {
+	netName, err := api.StateNetworkName(ctx)
+	if err != nil {
+		return nil, err
+	}
+	subnetID := address.SubnetID(netName)
+
+	var validators []hierarchical.Validator
+
+	minersEnv := os.Getenv(ValidatorsEnv)
```
matejpavlovic: Is this somehow different from validators or could the variable be named differently?

chain/consensus/mir/request_cache.go
```diff
@@ -0,0 +1,62 @@
...
+	cache          map[string]ClientRequest
+	handledClients map[string]bool
+}
+
+// addIfNotExist adds the request if key h doesn't exist .
+func (c *requestCache) addIfNotExist(clientID, h string, r Request) (exist bool) {
+	c.lk.Lock()
+	defer c.lk.Unlock()
+
+	if exist = c.handledClients[clientID]; !exist {
```
matejpavlovic, line 40: What if a request with a particular clientID is already stored in the cache, but now we are adding a different request from the same client. It looks like we will ignore the second one. Is that intended?

chain/consensus/mir/request_cache.go
```diff
@@ -0,0 +1,62 @@
...
+		c.cache[h] = ClientRequest{r, clientID}
+		c.handledClients[clientID] = true
+	}
+	return
+}
+
+// getDel gets the target request by the key k and deletes the keys.
+func (c *requestCache) getDel(h string) (Request, bool) {
```
matejpavlovic, line 48: In general, does the cache only support storing one request per client at a time?

chain/consensus/mir/request_cache.go
```diff
@@ -0,0 +1,62 @@
...
+	r, ok := c.cache[h]
+	if ok {
+		delete(c.handledClients, r.ClientID)
+		delete(c.cache, h)
+	}
+	return r.Request, ok
+}
+
+// getRequest gets the request by the key.
+func (c *requestCache) getRequestItem(k string) (Request, bool) {
```
matejpavlovic: Why `Item`? Could we simply call it `getRequest`? Also, should't we export that method?

matejpavlovic (changes requested):

chain/consensus/mir/manager.go
dnkolegov: We agreed with Alfonso to use xerrors as before

chain/consensus/mir/manager.go
dnkolegov: Done

chain/consensus/mir/manager.go
dnkolegov: done

chain/consensus/mir/manager.go
dnkolegov: done

chain/consensus/mir/app.go
dnkolegov: Eudico does need empty blocks as well at least for testing and demo purposes because otherwise, you will not be able to send a message. Deleted

chain/consensus/mir/request_cache.go
dnkolegov: You are right. I exposed this method via Manager

chain/consensus/mir/request_cache.go
dnkolegov: Yes, it does. All messages are stored in mempool, we don't add a message from mempool for the client if the cache already has a message for that client.

chain/consensus/mir/manager.go

chain/consensus/mir/request_cache.go
dnkolegov: Yes. That is intended.
So this cache stores only requests that have been sent to Mir nodes - one request per client.
Because that cache is just a PoC.

chain/consensus/mir/manager.go
dnkolegov: That's a common approach in Eudico and Lotus. So I just follow this. In each package logger is a global variable.
We can pass that package level logger into Manager's New, if needed

chain/consensus/mir/manager.go
dnkolegov: That how it looks like in logs
```
{"info": ["epoch",10,"instance",0,"view",0]}
```
Again, I use the approach from Eudico/Lotus.

chain/consensus/mir/manager.go
matejpavlovic: ok, it's not that important for now. When we need more instances of the manager, we can easily adapt the code.

PR #207, opened 2022-06-13
Add finality parameter for HC

dnkolegov:
This PR makes it possible to configure the finality threshold for a subnet via CLI.


api/api_hierarchical.go
```diff
@@ -14,29 +14,29 @@ import (
...
-	SubnetChainHead(context.Context, address.SubnetID) (*types.TipSet, error)                                                                                                                                                  // perm:read
-	SubnetStateGetActor(ctx context.Context, id address.SubnetID, addr address.Address, tsk types.TipSetKey) (*types.Actor, error)                                                                                             // perm:read
-	SubnetStateWaitMsg(ctx context.Context, id address.SubnetID, cid cid.Cid, confidence uint64, limit abi.ChainEpoch, allowReplaced bool) (*MsgLookup, error)                                                                 // perm:read
-	SubnetStateGetValidators(ctx context.Context, id address.SubnetID) ([]hierarchical.Validator, error)                                                                                                                       // perm:read
+	AddSubnet(ctx context.Context, addr address.Address, parent address.SubnetID, name string, cns uint64, stake abi.TokenAmount, checkPeriod abi.ChainEpoch, threshold abi.ChainEpoch, p *hierarchical.ConsensusParams) (address.Address, error) // perm:write
```
adlrocha: We can probably leave it as-is for now (unless you strongly feel it is worth changing it now), but we should consider aggregating `ConsensusParams`, and all these arguments into a `SubnetParams` struct or something to prevent the arguments of the functions from growing every time we need to add something new.

chain/consensus/hierarchical/actors/subnet/subnet_state.go
```diff
@@ -128,6 +130,15 @@ func ConstructSubnetState(store adt.Store, params *ConstructParams) (*SubnetStat
 		period = sca.DefaultCheckpointPeriod
 	}
 
+	finalityThreshold := params.FinalityThreshold
+	if finalityThreshold < 1 {
```
adlrocha: Is this correct? Why are you reassigning if `finalityThreshold < 1`? (you already assign it to the variable in the previous line)

cmd/eudico/subnet.go
```diff
@@ -100,27 +100,31 @@ var addCmd = &cli.Command{
...
-			Name:  "delegminer",
+			Name:  "delegated-miner",
 			Usage: "optionally specify miner for delegated consensus",
 		},
 		&cli.Uint64Flag{
 			Name:  "min-validators",
 			Usage: "optionally specify number of validators in subnet",
 			Value: 0,
 		},
+		&cli.IntFlag{
```
adlrocha, line 124: Ideally, I would say that even if we make this an optional argument, we should offer a reference value for each consensus algorithm implementation (or at least a lower bound, as with every consensus algorithm the required finality may be different)

adlrocha (approved):
LGTM. Before you merge, can you have a look at the comment about the `finalityThreshold` assignment? I don't know if it may be a slight bug. Thanks!

api/api_hierarchical.go
dnkolegov: Will do!

chain/consensus/hierarchical/actors/subnet/subnet_state.go
dnkolegov: You are right. It should be assigned to `subnetmgr.FinalityThreshold`

cmd/eudico/subnet.go
dnkolegov: Do we know the reference values for each consensus now? If we do I can add them. Probably we should do the same with `checkPeriod`?

adlrocha: Not yet, probably worth adding a `FIXME` for future work. We could come up with something for the current consensus algorithms but it'll probably not be accurate.

Good idea also for the `checkPeriod'

dnkolegov: Fixed

build/params_2k.go
```diff
@@ -120,3 +120,17 @@ const InteractivePoRepConfidence = 6
 const BootstrapPeerThreshold = 1
 
 var WhitelistedBlock = cid.Undef
+
+const DelegatedPoWFinality = 2
+const PoWFinality = 2
+const TendermintFinality = 1
+const MirFinality = 1
+const DummyFinality = 1
+const FilecoinECFinality = 2
```
adlrocha: `2` may be too low for Filecoin EC. Let's put 5 for now (although it is also quite low). Eudico builds from `2k` so even on testnet we would have this value. Not a pressing problem, we can change it in the future.

build/params_2k.go
```diff
@@ -120,3 +120,17 @@ const InteractivePoRepConfidence = 6
 const BootstrapPeerThreshold = 1
 
 var WhitelistedBlock = cid.Undef
+
+const DelegatedPoWFinality = 2
```
adlrocha: Why not using a map instead a list of consts in the end?

chain/consensus/hierarchical/actors/subnet/subnet_actor.go
```diff
@@ -459,15 +460,15 @@ func (st *SubnetState) addStake(rt runtime.Runtime, sourceAddr address.Address,
...
+	// Remove from stakes.
 	err := stakes.MustSubtract(sourceAddr, minerStake)
 	builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed subtracting ablanace for miner")
 	// Flush stakes adding miner stake.
 	st.Stake, err = stakes.Root()
 	builtin.RequireNoErr(rt, err, exitcode.ErrIllegalState, "failed to flust stakes")
 
 	// Remove miner from list of miners if it is there.
-	// NOTE: If we decide to support part-recovery of stake from stards
+	// NOTE: If we decide to support part-recovery of stake from shards
```
adlrocha:
```suggestion
	// NOTE: If we decide to support part-recovery of stake from subnets
```

adlrocha (approved):
LGTM. My only concern is that while reading the code `manager.Service` is not quite clear what is refers to. Can we add some indication that is related to subnets? Calling the package `submgr` instead of `manager` could do the trick, so we have `submgr.Service`. 

If you see this makes no sense feel free to merge, but I think this would make the code more readable.

build/params_2k.go
dnkolegov: We will get cycles.

build/params_2k.go
dnkolegov: fixed

chain/consensus/hierarchical/actors/subnet/subnet_actor.go
dnkolegov: fixed

PR #208, opened 2022-06-16
Mir crypto

dnkolegov:
This PR adds an implementation of Mir's Crypto interface.


chain/consensus/mir/crypto.go
```diff
@@ -0,0 +1,124 @@
...
+	}
+	return c.verifySig(data, signature, nodeAddr)
+}
+
+// RegisterNodeKey associates a public key with a numeric node ID.
+// The representation of the key is implementation-dependent.
+// Calls to VerifyNodeSig will fail until RegisterNodeKey is successfully called with the corresponding node ID.
+// Returns nil on success, a non-nil error on failure.
+func (c *CryptoManager) RegisterNodeKey(pubKey []byte, nodeID t.NodeID) error {
+	panic("not implemented")
```
adlrocha: Maybe I'm missing something but, why do we need to sign with the host identity and with a `walletID`? We should have look at https://github.com/libp2p/go-libp2p-core/tree/master/crypto, libp2p hosts already hold an identity and can sign with their libp2p identity.

chain/consensus/mir/crypto.go
```diff
@@ -0,0 +1,124 @@
...
+}
+
+// Sign signs the provided data and returns the resulting signature.
+// The data to be signed is the concatenation of all the passed byte slices.
+// A signature produced by Sign is verifiable using VerifyNodeSig or VerifyClientSig,
+// if, respectively, RegisterNodeKey or RegisterClientKey has been invoked with the corresponding public key.
+// Note that the private key used to produce the signature cannot be set ("registered") through this interface.
+// Storing and using the private key is completely implementation-dependent.
+func (c *CryptoManager) Sign(data [][]byte) (bytes []byte, err error) {
+	signature, err := c.wallet.WalletSign(context.Background(), c.addr, hash(data), MsgMeta)
```
adlrocha: Why use a plain hash instead of a `multihash` or a `cid`?

adlrocha (commented):
I guess this just a draft, added a few comments. Let me know if you need me to give it a deeper look.

chain/consensus/mir/crypto.go
matejpavlovic: In fact, the `RegisterNodeKey` (and other functions importing and deleting keys) are gone anyway in the newest restructured Mir implementation. It's really just about signing and verifying now.

chain/consensus/mir/crypto.go
matejpavlovic: Whatever we sign, it has to be computed locally, and I find the solution with the plain hash good for now, mostly because it is simple. Later, when we want to support multiple different implementation s of the crypto module, we'll need to specify exactly how the signature over the given data is computed, including what kind of hash to use (if any). As long all crypto modules return the same signature over the same data, we're good. And I'd say that for now it's ok to count on a single simple implementation.

chain/consensus/mir/crypto.go
```diff
@@ -35,33 +26,81 @@ type CryptoManager struct {
...
+// Any subsequent call to VerifyNodeSig(..., nodeID) will fail.
+func (c *CryptoManager) DeleteNodeKey(nodeID t.NodeID) {
+	panic("not implemented")
+}
+
+// DeleteClientKey removes the public key associated with clientID from the module's state.
+// Any subsequent call to VerifyClientSig(..., clientID) will fail.
+func (c *CryptoManager) DeleteClientKey(clientID t.ClientID) {
+	panic("not implemented")
+}
```
matejpavlovic: No need for these any more.

matejpavlovic (commented):
Nice!
Now that the upgrade to mir is finally merged, the only thing that needs to be implemented is the [`mirCrypto.Impl` interface](https://github.com/filecoin-project/mir/blob/main/pkg/crypto/api.go) of Mir.

Then we can use it as a module like this
```go
cryptoManager, err := NewCryptoManager(...)
// handle err...
cryptoModule := mirCrypto.New(cryptoManager)
// ...
// use cryptoModule when instantiating a Node with mir.NewNode(...)
```

matejpavlovic:
## Discussion

### Question

What is the `VerifyClientSig` function and do we need it? How is it different from `VerifyNodeSig`?

### Answer

It is meant for verifying signatures on transactions in an SMR system. In Eudico-terms, this would be for verifying signatures on Eudico messages that were produced using a wallet key.

On the other hand, the `VerifyNodeSig` function verifies signatures on some *protocol* messages. These signatures are created explicitly by the protocol implementation and the keys correspond to the validator roles of the nodes. In Eudico, these would be the keys that define the libp2p identities of nodes.

Given that in Eudico, the mempool already only contains Eudico messages that are ready to be ordered and their signatures do not need to be verified by the ordering layer, the `VerifyClientSig` function of the crypto module is not required and can be removed from the module.

### Question

What key will we use for signing consensus protocol messages?

### Answer

Ideally we should use the key that corresponds to the identity of the validator node as defined in the validator set.
In other words, whatever is used in whatever representation of the validator set to identify a validator, it must be (explicitly or implicitly) linked to a key pair such that
1) the validator can use the private part to sign protocol messages and
2) other validators can use the public part for verifying those messages.

## Conclusion

### Options

We have two options for implementing the crypto module for Eudico

1) Use the libp2p-based identity of nodes and sign/verify protocol messages using the associated key pair. While this is the cleanest solution, it might have a practical caveat that the libp2p key (or an interface for signing arbitrary data with it) might not be easily accessible to the Mir implementation.

2) Use the libp2p-based identity of nodes only to identify them on the network and explicitly associate a wallet key/address with this identity in the membership configuration.

### How to proceed

Look for possibilities of signing arbitrary data using the libp2p key, but without spending much time on it (few hours max). On success, use option 1. On failure, use option 2.

chain/consensus/mir/crypto.go
dnkolegov: fixed

chain/consensus/mir/crypto_test.go
```diff
@@ -48,16 +51,43 @@ func (n *cryptoNode) WalletSign(ctx context.Context, k address.Address, msg []by
...
 }
 
+// ---------------
+
+type cryptoNode struct {
+	privKey libp2pcrypto.PrivKey
+	id      peer.ID
+}
+
+func newCryptoNode() (*cryptoNode, error) {
```
adlrocha: So we are using a new identity for Mir instead of using the one from our `libp2p.Host` right? I guess this is OK for now, we can always change it once we integrate the transport layer.

chain/consensus/mir/crypto.go
```diff
@@ -16,22 +18,33 @@ import (
...
 }
 
-type WalletCrypto interface {
-	WalletSign(ctx context.Context, k address.Address, msg []byte) (*filcrypto.Signature, error)
-	WalletVerify(ctx context.Context, k address.Address, msg []byte, sig *filcrypto.Signature) (bool, error)
+// CryptoManager implements Mir's crypto interface.
+// It uses the node's libp2p private key to sign Mir messages.
+// libp2p public keys and IDs correspond to each other.
+// Because of that we don't need trust establishment mechanisms like CA.
+// We ED25519 signing algorithm, because Filecoin uses it.
```
matejpavlovic: We use the ...

chain/consensus/mir/crypto_test.go
dnkolegov: We will use the real node ID retrieved via `host.ID()`. `newCryptoNode` is a dummy implementation of the interface to not inject the real full node API in unit tests.

matejpavlovic (commented):
LGTM.
I'll adapt the Mir crypto interface so we can simplify even further.

chain/consensus/mir/crypto.go
```diff
@@ -0,0 +1,92 @@
...
+var _ mircrypto.Crypto = &CryptoManager{}
+
+type CryptoManager struct {
+	key address.Address // The address corresponding to the private key.
+	api WalletCrypto    // API used to sign data in HSM-model.
+}
+
+func NewCryptoManager(key address.Address, wallet WalletCrypto) (*CryptoManager, error) {
+	if key.Protocol() != address.SECP256K1 {
+		return nil, xerrors.New("must be SECP address")
```
matejpavlovic: `fmt.Errorf`?

chain/consensus/mir/crypto.go
```diff
@@ -0,0 +1,92 @@
...
+	for _, d := range data {
+		h.Write(d)
+	}
+	return h.Sum(nil)
+}
+
+func getAddr(nodeID string) (address.Address, error) {
+	addrParts := strings.Split(nodeID, ":")
+	if len(addrParts) != 2 {
+		return address.Undef, xerrors.Errorf("invalid node ID: %s", nodeID)
```
matejpavlovic: `fmt.Errorf`?

chain/consensus/mir/manager.go
```diff
@@ -135,30 +137,27 @@ func NewManager(ctx context.Context, addr address.Address, api v1api.FullNode) (
...
-			RequestStore: reqStore,
-			Protocol:     issProtocol,
-			App:          app,
-			Crypto:       &mirCrypto.DummyCrypto{DummySig: []byte{0}},
-		})
+		map[t.ModuleID]modules.Module{
+			"net":    net,
+			"iss":    issProtocol,
+			"app":    app,
+			"crypto": mirCrypto.New(&mirCrypto.DummyCrypto{DummySig: []byte{0}}),
```
matejpavlovic: Can we already use the new crypto implementation here?

chain/consensus/mir/manager.go
```diff
@@ -182,23 +181,26 @@ func NewManager(ctx context.Context, addr address.Address, api v1api.FullNode) (
...
 	errChan := make(chan error, 1)
-	managerCtx, managerCancel := context.WithCancel(ctx)
+	managerCtx, managerCancel := context.WithCancel(context.Background())
 
 	go func() {
 		select {
 		case <-ctx.Done():
-			log.Debugf("Mir manager: context closed")
+			log.Infof("Mir manager %s: context closed", m.MirID)
 			m.Stop()
```
matejpavlovic: Is the manager supposed to stop only when the context given to it from outside is canceled, or should it also stop when the Mir node stops by itself for some reason (e.g. due to an error)?
Currently, if the Mir node encounters an error, the Manager is not stopped.
Moreover, if the Mir node stops by itself without an error, this whole goroutine will keep running until the external context is canceled.
Is that intended?

chain/consensus/mir/manager.go
```diff
@@ -218,8 +220,13 @@ func (m *Manager) Stop() {
 
 func (m *Manager) SubmitRequests(ctx context.Context, refs []*RequestRef) {
 	for _, r := range refs {
-		err := m.MirNode.SubmitRequest(ctx, r.ClientID, r.ReqNo, r.Hash, []byte{0})
-		if err != nil {
+
+		e := events.NewClientRequests("iss",
+			[]*requestpb.Request{
```
matejpavlovic: Now that the `NewClientRequests` function accepts a whole list of requests for a single event, it would be better to first create a slice of all the client requests and then only create a single `NewClientRequests` event and inject only that one event.

chain/consensus/mir/mine.go
```diff
@@ -69,7 +69,7 @@ func Mine(ctx context.Context, addr address.Address, api v1api.FullNode) error {
 			log.Debug("Mir miner: context closed")
 			return nil
 		case err := <-mirErrors:
-			return err
+			return xerrors.Errorf("Mir miner consensus error", err)
```
matejpavlovic: `fmt.Errorf`?

matejpavlovic (commented):
Gooks good in general, only minor comments, the most important one being the usage of the new crypto module.

We'll need to see about including the request payload, but that's for another PR.

chain/consensus/mir/manager.go
```diff
@@ -145,6 +141,11 @@ func NewManager(ctx context.Context, addr address.Address, api v1api.FullNode) (
 	}
 
 	app := NewApplication()
+	pool := newRequestPool()
+	cryptoManager, err := NewCryptoManager(addr, api)
+	if err != nil {
+		return nil, xerrors.Errorf("failed to create crypto manager: %w", err)
```
matejpavlovic: `fmt.Errorf`?

chain/consensus/mir/crypto.go
dnkolegov: ok

chain/consensus/mir/crypto.go
dnkolegov: ok

chain/consensus/mir/manager.go
dnkolegov: Fixed

chain/consensus/mir/mine.go
dnkolegov: ok

chain/consensus/mir/manager.go
dnkolegov: ok

chain/consensus/mir/manager.go

chain/consensus/mir/manager.go
dnkolegov: 1. Case 1. If the Mir node encounters an error then we send the error to `errChan` and call `managerCancel()` that will stop the manager as a goroutine but it will not call the `Stop` function. `Stop` doesn't stop the manager it stops the Mir node. The question: if the Mir node has encountered an error does make sense to stop its components gracefully?
2. Case 2. `Mir node stops by itself without an error`. According to the comments it `Stops and returns when ctx is canceled.` So it can't just finish all work and stop. Right? It stops via context only and then the special error is returned.

matejpavlovic: Then I don't quite understand it... In the code it looks like `m` is the Manager and the code calls `m.Stop()`. Doesn't that stop the Manager?

dnkolegov:
m.Stop() stops the Mir network and WAL.


вт, 21 июн. 2022 г., 13:22 Matej Pavlovic ***@***.***>:

chain/consensus/mir/manager.go
matejpavlovic: > if the Mir node has encountered an error does make sense to stop its components gracefully?

Yes. The Mir Node does not really have a reason (or even a way) of stopping other modules like the WAL or Net on error. So in fact, they still need to be stopped from the outside event if the Node fails. In a way, the modules can be seen as components of the Node, but one can also see them as independent components that only interact using the node (after all they are also started separately from the Node). So it make sense to also stop them (preferably gracefully) separately from the Node (even if it fails).

It also happens to simplify the code to something like this:
```go
func (m *Manager) Start(ctx context.Context) chan error {
	log.Infof("Mir manager %s starting", m.MirID)

	errChan := make(chan error, 1)

	go func() {
		
		// Run Mir node until it stops.
		if err := m.MirNode.Run(ctx); err != nil && !errors.Is(err, mir.ErrStopped) {
			log.Infof("Mir manager %s: Mir node stopped with error: %v", m.MirID, err)
			errChan <- err
		} else {
			log.Infof("Mir manager %s: Mir node stopped gracefully", m.MirID)
			errChan <- nil
		}
		
		// Perform cleanup of Node's modules.
		m.Stop()
	}()

	return errChan
}
```

PR #214, opened 2022-06-24
New Eudico ordering interface

dnkolegov:
This PR is an implementation of the interface - https://hackmd.io/6v00B72USpav6rDUUhqxAw - that was updated recently

### DIFF
--- a/chain/consensus/mir/mine.go
+++ b/chain/consensus/mir/mine.go
@@ -11,7 +11,6 @@ import (
 	"github.com/filecoin-project/lotus/chain/consensus/platform/logging"
 	"github.com/filecoin-project/lotus/chain/types"
 	ltypes "github.com/filecoin-project/lotus/chain/types"
-	mirRequest "github.com/filecoin-project/mir/pkg/pb/requestpb"
 )
 
 // Mine handles block "mining" using Mir framework.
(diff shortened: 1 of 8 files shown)


chain/consensus/mir/mine.go
```diff
@@ -128,21 +129,21 @@ func Mine(ctx context.Context, addr address.Address, api v1api.FullNode) error {
...
 			}
 
-			refs, err = m.AddCrossMessages(refs, crossMsgs)
+			requests, err = m.AddCrossMessages(requests, crossMsgs)
 			if err != nil {
 				log.With("epoch", nextEpoch).
 					Errorw("unable to push cross-messages", "error", err)
 			}
 
-			m.SubmitRequests(ctx, refs)
```
matejpavlovic, line 145: I realized that this is probably called a lot, as it is in the `default` branch of the `select` statement. It looks like it creates a sort of busy-wait loop, always burning up one CPU no matter what.
It's not a big deal for now though, as I expect this branch to be part of the request pool (mempool wrapper) module, where it we'll take this into account.

This is only a comment, no direct action required, so pls mark it directly as resolved yourself when you read it.

matejpavlovic (approved):

PR #215, opened 2022-06-27
Propogate checkPeriod

dnkolegov:


chain/consensus/hierarchical/actors/subnet/genesis.go
```diff
@@ -199,9 +199,9 @@ func CreateGenesisFile(
 
 	t := uint64(time.Now().Unix())
 
-	checkPeriod := consensus.ConsensusCheckPeriod(cns)
+	chp := consensus.DefaultCheckpointPeriod(alg)
```
adlrocha: I don't know if we are setting always to the defualt value the checkpoint period in the SCA. We are not passing any values here, and `WriteGenesis` is the one responsible for setting the checkpoint period of the SCA. Apart from that I think everything looks good.

adlrocha (approved):
LGTM. Before merging please have a quick look at my comment (I don't know if it may be a bug).

chain/consensus/hierarchical/actors/subnet/genesis.go
dnkolegov: `CreateGenesisFunction` you are commenting here is used only to run consensus in root subnets.
As we discussed in the Slack: 
- dnkolegov: `The next question: do you agree that if we run consensus (non-FilecoinEC) in the root then we should use the default checkpoint period? Because overwise we have to add a check period parameter to genesis generation commands`
- adlrocha: `Yes, that's fine. It's actually not used for anything`

But if we run consensus in a subnet then `WriteGenesis` is used and that function is parametrized by the `CheckpointPeriod` from subnet parameters:

https://github.com/filecoin-project/eudico/blob/91ce6604318793527812795b221b47aacc16459e/chain/consensus/hierarchical/actors/subnet/subnet_actor.go#L103

Let me know if I missed something.

dnkolegov: But I found the following bug: checkpoint period and finality supplied by users are checked in `ConstructSubnetState` but then the original parameters are passed to the `initGenesis` function. Did you mean this?

Moreover, now, if the checkpoint period is less than the default checkpoint period for the consensus then we will use the default value.  Let's return an error if checkpoint period or finality threshold values provided by users are wrong. 
Let's change the logic. Do you agree?

dnkolegov: The draft:
```
        if params.CheckpointPeriod == 0 {
		params.CheckpointPeriod = hierarchical.DefaultCheckpointPeriod(params.Consensus)
	}

	if params.FinalityThreshold == 0 {
		params.FinalityThreshold = hierarchical.DefaultFinality(params.Consensus)
	}

	if params.CheckpointPeriod < hierarchical.DefaultCheckpointPeriod(params.Consensus) {
		return nil, xerrors.Errorf("checkpoint value is too low (%v)", params.CheckpointPeriod)
	}
	if params.FinalityThreshold < hierarchical.DefaultFinality(params.Consensus) {
		return nil, xerrors.Errorf("finality value is too low (%v)", params.FinalityThreshold)
	}

	// Finality should always be less than the checkpoint period.
	if params.FinalityThreshold >= params.CheckpointPeriod {
		return nil, xerrors.Errorf("finality threshold (%v) must be less than checkpoint period (%v)",
			params.FinalityThreshold, params.CheckpointPeriod)
	}

	parentID := address.SubnetID(params.NetworkName)

	return &SubnetState{
		ParentID:          parentID,
		Consensus:         params.Consensus,
		MinMinerStake:     params.MinMinerStake,
		Miners:            make([]address.Address, 0),
		Stake:             emptyStakeCid,
		Status:            Instantiated,
		CheckPeriod:       params.CheckpointPeriod,
		Checkpoints:       emptyCheckpointsMapCid,
		FinalityThreshold: params.FinalityThreshold,
		WindowChecks:      emptyWindowChecks,
		ValidatorSet:      make([]hierarchical.Validator, 0),
		MinValidators:     params.ConsensusParams.MinValidators,
	}, nil
```

adlrocha: I thought that `CreateGenesisFile` was also being used by subnets, but I see is just used for rootnets. For subnets we use directly `WriteGenesis`: https://github.com/filecoin-project/eudico/blob/c5fcf081e5385b5c77872b96e8a8758730a634e6/chain/consensus/hierarchical/actors/subnet/subnet_actor.go#L103 so I think it's correct.

Regarding the error, I think it is a good idea to return an error. It makes more sense that to return the default checkpoint period.

An additional note that I realized while reviewing this code. All of this will potentially change as we move this logic to FVM. So we'll need to review that everything makes sense there too.

dnkolegov: Fixed

PR #219, opened 2022-07-06
Removing cross-messages from blocks

### LINKED ISSUE #218: Remove `CrossMessage` field from `BlockMsg`
While writing the spec, we realized the `CrossMessage` field is no longer needed in `BlockMsg`. Before having hierarchical addresses we didn't have a way to identify cross-net messages, and we had to differentiate them from plain subnet messages because they are executed differently.
https://github.com/filecoin-project/eudico/blob/137c6ddc3c23e0ab776ec885c26c48c5b60b28ae/chain/types/blockmsg.go#L13

We can now identify cross-net messages in a block by simply looking at the `From`and `To` and checking if they are hierarchical addresses. Cross-net messages can be included in the `SecpkMessages` field of the `BlockMsg`.

This removes the need for backward-compatibility for `BlockMsgs` pre-hierarchical consensus (as explored in https://github.com/filecoin-project/eudico/pull/95)

adlrocha:
Fixes:  https://github.com/filecoin-project/eudico/issues/218
- It includes cross-messages inside the Secp message field of blocks with an `unknown` signature type.

### DIFF
--- a/chain/consensus/common/mine.go
+++ b/chain/consensus/common/mine.go
@@ -66,8 +66,8 @@ func PrepareBlockForSignature(ctx context.Context, sm *stmgr.StateManager, bt *l
 	}
 
 	for _, msg := range bt.CrossMessages {
-		m := &types.SignedMessage{Message: *msg, Signature: crypto.Signature{Type: crypto.SigTypeUnknown}}
-		c, err := sm.ChainStore().PutMessage(ctx, msg)
+		m := &types.SignedMessage{Message: *msg, Signature: crypto.Signature{Type: crypto.SigTypeSecp256k1}}
+		c, err := sm.ChainStore().PutMessage(ctx, m)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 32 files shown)


chain/consensus/common/mine.go
```diff
@@ -66,12 +66,13 @@ func PrepareBlockForSignature(ctx context.Context, sm *stmgr.StateManager, bt *l
 	}
 
 	for _, msg := range bt.CrossMessages {
```
dnkolegov, line 68: Why block template does still have CrossMessages?

dnkolegov (approved):
LGTM.
But now we must make tests work again!

chain/consensus/common/mine.go
adlrocha: We are just keeping it in the block template for miners to be able to discern and prepare the cross-messages to include in the block (in case that's needed), but this doesn't affect the backward compatibility of the protocol.

PR #223, opened 2022-07-20
Mir Libp2p integration

dnkolegov:
This PR adopts Mir's Libp2p transport to Eudico and disables GRPC transport.


api/api_net.go
```diff
@@ -64,6 +64,9 @@ type Net interface {
 
 	// ID returns peerID of libp2p node backing this API
 	ID(context.Context) (peer.ID, error) //perm:read
+	// PrivKey returns libp2p host's private key.
+	// This is temporal experimental function that must be removed in the future.
```
adlrocha:
```suggestion
	// FIXME: This is temporal experimental function that must be removed in the future.
```
That way is easier to spot in the future? Exposing the private key is always a bit frightening :)

scripts/mir/connect.sh
```diff
@@ -0,0 +1,29 @@
+# This script is a helper connecting Eudico nodes with each other.
```
adlrocha, line 1: This is a really nice script :) It can really help testing and setting up networks.

node/impl/net/net.go
```diff
@@ -41,6 +42,12 @@ func (a *NetAPI) ID(context.Context) (peer.ID, error) {
 	return a.Host.ID(), nil
 }
 
+func (a *NetAPI) PrivKey(context.Context) ([]byte, error) {
```
adlrocha, line 50:
```suggestion
// FIXME: We should find a better way to pass the libp2p host to Mir
// without exposing the private key. 
func (a *NetAPI) PrivKey(context.Context) ([]byte, error) {
```

adlrocha: Now that I think about it, this may not be such a bad idea if we add a passphrase to encrypt the key. Kind of how Ethereum uses `unlockAccount` to unlock the private keys for signing.

chain/consensus/mir/utils.go
```diff
@@ -0,0 +1,78 @@
...
+		}
+		validators, err = api.SubnetStateGetValidators(ctx, subnetID)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get validators from state")
+		}
+	}
+	return validators, nil
+}
+
+func GRPCValidatorsMembership(validators []hierarchical.Validator) ([]t.NodeID, map[t.NodeID]string, error) {
```
adlrocha: I can't see where this function is used. Is this check still necessary without GRPC? Or do we still support GRPC within Eudico?

adlrocha (approved):
Awesome job! LGTM

chain/consensus/mir/utils.go
dnkolegov: You are right. This function was used to support GRPC connections. Removed

PR #225, opened 2022-07-21
fix atomic exec flaky test

### LINKED ISSUE #183: Test coverage for atomic execution protocol
The test coverage for HC's atomic execution needs to be improved. We currently only have a few unit tests for the offline execution, and the unit tests for the SCA. This issue will be fixed when:
- [ ] We have at least an itest test case for a "happy-path" of the atomic execution protocol.
- [ ] We fix the flaky tests from `sca_atomic.go`: https://github.com/filecoin-project/eudico/blob/eudico/chain/consensus/hierarchical/actors/sca/sca_atomic_test.go

adlrocha:
Fixes #183

Note: We may still need to have a more extensive coverage of the atomic execution tests, but as we are rewriting the actors in Rust, we'll defer that to the new implementation.


dnkolegov (approved):
LGTM! Great job! Now we will have more clear CI/CD pipeline!

PR #226, opened 2022-07-24
WIP - feat:metric:Introduce subnet related metrics

cryptoAtwill:
## Proposed Changes

- Adding `SubnetSpinUpDuration` and `SubnetCount` to the existing list of metrics
- Adding the measurements to the `addSubnet` command. 

## Additional Info

This PR is for early preview. Currently `lotus-stats` is giving some errors, need to find out why on that one first.

## Checklist

- [x] All commits have a clear commit message.
- [x] The PR title is in the form of of `<PR type>: <area>: <change being made>`
- [ ] CI is green


cmd/eudico/subnet.go
```diff
@@ -196,11 +200,21 @@ var addCmd = &cli.Command{
...
+
 		actorAddr, err := api.AddSubnet(ctx, params)
 		if err != nil {
 			return err
 		}
 
+		// we only record the successful calls
+		ms := time.Now().Sub(start).Microseconds()
+
+		sctx, _ := tag.New(ctx, tag.Upsert(metrics.SubnetType, cctx.String("consensus")))
```
adlrocha, line 214: This is not correct. To deploy a subnet, you first `subnet add` to deploy the subnet actor that will govern the subnet in the parent (no subnet is deployed in this step), and then when you `subnet join` is when all the stack for your subnet is initialized. Here you are only measuring the time to deploy the subnet actor.

Also, this is a sync command that waits for 5 epochs "finality" to be sure that the actor has been deployed successfully. I would suggest getting the metrics directly from the code at `chain/consensus/hierarchical`. It will allow you to get more fine-grained metrics (these commands abstract many of the low-level details of the protocol).

cmd/lotus-stats/README.md
```diff
@@ -7,7 +7,7 @@
 Influx configuration can be configured through env variables.
 
 ```
-LOTUS_STATS_INFLUX_ADDR="http://localhost:8086"
+LOTUS_STATS_INFLUX_ADDR="http://192.168.1.94:8086"
```
adlrocha, line 10: Why this IP? Is this the specific one that you got in your environment? If this is the case is probably better to leave it as `localhost` or a placeholder like `<influx_ip>`.

metrics/metrics.go
```diff
@@ -599,6 +606,17 @@ var (
...
 		Aggregation: view.Count(),
 	}
+
+	// hc
+	SubnetSpinUpDurationView = &view.View{
+		Measure:     SubnetSpinUpDuration,
+		Aggregation: defaultMillisecondsDistribution,
+		TagKeys:     []tag.Key{SubnetType},
+	}
+	SubnetCountView = &view.View{
```
adlrocha, line 616: If we add this metric, we should ensure that we count the active subnets (we increase it when we register a new subnet, and we decrease when we leave it). Actually, I think it'll be even more interesting to track the number of subnet in each state `Active`, `Inactive` and `killed`.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,38 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the README at:
+// https://github.com/microsoft/vscode-dev-containers/tree/v0.187.0/containers/go
```
adlrocha, line 2: What is this docker container? It'll be helpful to add a few notes on what is this, how it is used, etc.

adlrocha (changes requested):
I did a first pass. Ping if you need any additional reviews. Thanks!

cmd/eudico/subnet.go
cryptoAtwill: For add, do we measure the time here or actually in the join? We should measure the time when the msg gets back? Like at this point: https://github.com/filecoin-project/eudico/blob/110c4f2c8edceb2b8bb4489b4847c190a1dbf344/chain/consensus/hierarchical/subnet/submgr/manager.go#L532

adlrocha: As discussed in Slack, is really up to what you want to measure. Let's first gather a few sample metrics and then refine with different measurements of the end-to-end flow. Thanks!

adlrocha:
@cryptoAtwill, is this PR overwritten by #227? Should we close this one?

PR #227, opened 2022-07-26
add prometheus and remove influxdb

cryptoAtwill:
## Related Issues
Introduce `eudico-stats` with prometheus.

## Proposed Changes

- Add eudico-stats command to expose the metrics
- Add prometheus support
- Remove influxdb for now

## Additional Info

NA

## Checklist

- [ ] All commits have a clear commit message.
- [ ] The PR title is in the form of of `<PR type>: <area>: <change being made>`
- [ ] CI is green


Dockerfile.eudico
```diff
@@ -0,0 +1,6 @@
+FROM golang:1.18.3-buster
```
adlrocha: I don't get this image, what do we use it for?

cmd/eudico-stats/main.go
```diff
@@ -0,0 +1,138 @@
...
+
+	app := &cli.App{
+		Name:    "eudico-stats",
+		Usage:   "Collect basic information about a filecoin network using lotus",
+		Version: build.UserVersion(),
+		Flags: []cli.Flag{
+			&cli.StringFlag{
+				Name:    "lotus-path",
+				EnvVars: []string{"LOTUS_PATH"},
+				Value:   "~/.lotus", // TODO: Consider XDG_DATA_HOME
```
adlrocha: Why `XDG_DATA_HOME`?

.gitignore
```diff
@@ -53,4 +53,7 @@ bin/tmp/*
 scratchpad
 gen.gen
 
-**/*-wal/
\ No newline at end of file
+**/*-wal/
+.devcontainer
+/eudico-stats
```
adlrocha, line 58: This is because we are collecting the metrics in the current path by default?

cmd/eudico-stats/main.go
```diff
@@ -0,0 +1,138 @@
+package main
```
adlrocha, line 1: This file is a copy-past of lotus_stats, right? Are we collecting any metrics from Lotus' default views?

Makefile
```diff
@@ -225,6 +225,12 @@ lotus-stats:
 .PHONY: lotus-stats
 BINS+=lotus-stats
 
+eudico-stats:
+	rm -f eudico-stats
```
adlrocha, line 229: What if someone wants to keep their metrics, will re-building `eudic-stats` remove everything? Maybe consider including this in `make clean`?

adlrocha (commented):
Overall LGTM. Just a few comments to understand a bit better what you are doing, and I think we can merge it so you can have the core to start collecting HC metrics.

Dockerfile.eudico
cryptoAtwill: I was using this as the base image for vscode remote docker development. I will add more commands to it when `eudico-stats` is ready. The plan is to replace the Dockerfile.lotus for eudico.

cmd/eudico-stats/main.go
cryptoAtwill: I copied from the original code. Maybe I should just remove it.

.gitignore
cryptoAtwill: `/eudico-stats` is the built executable, haha. I just excluded it from being pushed to git.

cmd/eudico-stats/main.go
cryptoAtwill: Should be. Based on my current understanding, as long as the `stats.Record` is invoked, it is exported to prometheus. I also didnt see the Lotus default view in the `lotus-stats`, I'll check again when I add the HC metrics.

Makefile
cryptoAtwill: this should just remove the command. I think in the future, the data should be persisted in the db, so building or removing should not affect. And once the db is in place, I will add back the `reset` flag so that ppl can clean the existing data.

.gitignore
adlrocha: Ah, understood. LGTM.

cmd/eudico-stats/main.go
adlrocha: Maybe remove the comments (is confusing) and change `LOTUS_PATH` for `EUDICO_PATH` and use `~/.eudico` for the default location. This is the case for eudico

Dockerfile.eudico
adlrocha: Let's not replace it for now, but let's have a Dockerfile.lotus and Dockerfile.eudico. We are still a fork, and we want to be able to pull potential changes they make upstream. 

Let's have a minimal dockerfile working in this PR for Eudico or let's remove it for now (this is a raw Go image without much).

cryptoAtwill: I'll remove for now.

.gitignore
```diff
@@ -53,4 +53,8 @@ bin/tmp/*
...
+f1*
```
adlrocha: Maybe ignore `*.key` better? There are also `f2` addresses.

.gitignore
```diff
@@ -53,4 +53,8 @@ bin/tmp/*
...
+Dockerfile.eudico
```
adlrocha: Why do we ignore this? I don't expect anyone to have this file locally (except maybe you in your working environment). I'd rather remove this from the .gitignore.

adlrocha (approved):
LGTM. Some minor fixes before we merge in the .gitignore. Also, I don't see the daemon change you showed me sync, is that coming in some other PR? Thanks!

PR #233, opened 2022-07-28
Features/add initial subnet metrics

cryptoAtwill:
## Proposed Changes

Add initial metrics to track:
- [x] Number of created subnets
- [x] Number of killed subnets 
- [x] Number of active subnets
- [x] Grafana dashboard
- [x] Documentation and scripts to setup

Sample dashboard:
image

## Checklist

- [ ] All commits have a clear commit message.
- [ ] The PR title is in the form of of `<PR type>: <area>: <change being made>`
- [ ] CI is green


chain/consensus/hierarchical/subnet/submgr/manager.go
```diff
@@ -102,7 +104,7 @@ type SubnetParams struct {
 func NewService(
 	mctx helpers.MetricsCtx,
 	lc fx.Lifecycle,
-	// api impl.FullNodeAPI,
+// api impl.FullNodeAPI,
```
adlrocha:
```suggestion
```

chain/consensus/hierarchical/subnet/submgr/manager.go
```diff
@@ -455,6 +462,8 @@ func (s *Service) AddSubnet(ctx context.Context, params *hierarchical.SubnetPara
 		return address.Undef, aerr
 	}
 
+	stats.Record(ctx, metrics.SubnetCreatedCount.M(1))
```
adlrocha, line 465: I don't think this is is right. Instead of counting every time we create a new subnet, I would keep polling the state periodically (as you do below) and update the metrics "real-time" accordingly. Directly from the state we can get the number of subnets created, active an inactive, no need to count here.

adlrocha: Unless we want to track the total amount of subnets created in the history of the subnet, in which case this is correct :)

chain/consensus/hierarchical/subnet/submgr/manager.go
```diff
@@ -548,6 +559,28 @@ func (s *Service) JoinSubnet(
...
+		pbs := blockstore.NewAPIBlockstore(parentAPI)
+		pcst := cbor.NewCborStore(pbs)
+		if err := pcst.Get(ctx, subnetAct.Head, &newst); err != nil {
+			log.Infow("getting actor state error after join: %w", err)
+			return
+		}
+
+		if (startStatus == subnet.Inactive || startStatus == subnet.Instantiated) &&
+			newst.Status == subnet.Active {
+			stats.Record(ctx, metrics.SubnetActiveCount.M(1))
```
adlrocha: I think we really need the EventListener to do this in the right way. Instead of having to count up and down as a peer joins or leave a subnet, we could report the absolute number of active and killed subnets. Even if this approach is easier, I think it is worth exploring the event listener approach right away as it will give us all that we need to report certain metrics "real-time".

cmd/eudico/daemon.go
```diff
@@ -193,6 +194,20 @@ func daemonCmd(overrides node.Option) *cli.Command {
 			// Set the metric to one so it is published to the exporter
 			stats.Record(ctx, metrics.LotusInfo.M(1))
 
+			go func() {
+				exporter, err := prometheus.NewExporter(prometheus.Options{
+					Namespace: "lotus_stats",
```
adlrocha:
```suggestion
					Namespace: "eudico_stats",
```

cmd/eudico/daemon.go
```diff
@@ -193,6 +194,20 @@ func daemonCmd(overrides node.Option) *cli.Command {
 			// Set the metric to one so it is published to the exporter
 			stats.Record(ctx, metrics.LotusInfo.M(1))
 
+			go func() {
```
adlrocha: We should probably extract this into its own command. We could start this in an independent process, spawn the required APIs from the repo, and spawn all the metrics-related processes that we need.

Either that or this should be opt-in through an env variable.

metrics/metrics.go
```diff
@@ -53,6 +53,10 @@ var (
 	ProtocolID, _ = tag.NewKey("proto")
 	Direction, _  = tag.NewKey("direction")
 	UseFD, _      = tag.NewKey("use_fd")
+
+	// Subnet ID
+	SubnetID, _    = tag.NewKey("subnet_id")
+	SubnetLevel, _ = tag.NewKey("subnet_level") // The subnet level, the number of parents it has
```
adlrocha: Number of ancestors? Subnets can only have one parent

chain/consensus/hierarchical/subnet/submgr/manager.go
cryptoAtwill: Yeah, I was thinking about how many times add subnet is invoked in history.

metrics/metrics.go
cryptoAtwill: Number of ancestors seems to be the more correct term

adlrocha:
We are making good progress. Thanks! I would try the event listener approach as part of this PR now that you know how to collect metrics. I think it can give all the skeleton that we need to collect "real-time" metrics.

cmd/eudico/daemon.go
cryptoAtwill: I'll add an env variable for now.

cmd/eudico/daemon.go
```diff
@@ -193,6 +199,22 @@ func daemonCmd(overrides node.Option) *cli.Command {
 			// Set the metric to one so it is published to the exporter
 			stats.Record(ctx, metrics.LotusInfo.M(1))
 
+			if cctx.Bool("export-metrics") {
```
adlrocha, line 204: This looks good, just a note to remind ourselves to consider this in an independent process. 

```suggestion
                        // FIXME: Consider moving this to an independent optional process
                        // (so that is not dependent on the main process).
			if cctx.Bool("export-metrics") {
```

cmd/eudico-stats/main.go
```diff
@@ -115,24 +124,54 @@ var runCmd = &cli.Command{
 		genesisTime := time.Unix(int64(gtp.MinTimestamp()), 0)
 
 		go func() {
-			t := time.NewTicker(time.Second)
+			// trigger calculation every 30 seconds
```
adlrocha: It is actually doing it every second right?

This approach is OK for the current metric, but we should consider the `EventListener` approach so that we have fine-grained data for others.

adlrocha (commented):
So far this looks good. What about adding an additional metric to test if the 1 second approach works fine for other "less permanent" metrics? Thanks!

cmd/eudico-stats/main.go
cryptoAtwill: I can delay this to every 30 seconds.
Yeah, I tried setup the event listener, but probably something I did wrongly. But the `lotus-stats` is also doing sth similar for some of the metrics I think.

adlrocha: This is fine, I just asked because the comment and the code don't match.

adlrocha: And the suggestion of collecting additional metrics is to get a sense of the processes we need in place before committing a lot of time to an approach that doesn't work and having to revert this after merging.

chain/consensus/hierarchical/actors/sca/sca_actor.go
```diff
@@ -781,3 +787,15 @@ func (a SubnetCoordActor) SubmitAtomicExec(rt runtime.Runtime, params *SubmitExe
 	// Return status of the execution
 	return &SubmitOutput{exec.Status}
 }
+
+func recordCrossNetMsgSend(rt runtime.Runtime, msg types.Message, from string, to string, msgType string) {
```
adlrocha: This is actor code and runs inside the VM, I don't think you want to track this from the actor code. Even more, this code will disappear soon, as we will remove all LegacyVM code in favor of the FVM, so you'll have to track this from the state and not from its execution.

chain/consensus/hierarchical/actors/sca/sca_apply.go
```diff
@@ -190,3 +195,23 @@ func noopWithStateTransaction(rt runtime.Runtime, msg types.Message, code exitco
 		noop(rt, &st, msg, code, err)
 	})
 }
+
+func recordCrossNetMsgExecution(rt runtime.Runtime, msg types.Message, code exitcode.ExitCode, msgType string) {
```
adlrocha: Same here, this is executed by every peer in the network, you need to find a better way to track these messages without coupling new code in the actor. 

A good way of doing so is through the event listener. You can check `TopDownMsg` and `BottomUpMsg` variables and see if there are new messages to be executed.

chain/consensus/hierarchical/subnet/submgr/manager.go
adlrocha, line 465: It's probably better to collect this metric from the state, what if adding the subnet fails and we record as if it was created when it's not really the case?

cmd/eudico-stats/README.md
```diff
@@ -23,5 +23,19 @@ cd cmd/eudico-stats && docker-compose up -d && cd -
...
+./eudico wallet set-default <Your Key>
+./eudico subnet add --name test1 --consensus POW
+./eudico subnet join --subnet /root/t01000 4
 # you should be able to see the stats on http://localhost:9090 with eudico*
 ```
+
+Using docker-compose
+In the root directory
+```shell
+docker build -t eudico:latest -f Dockerfile.eudico .
```
adlrocha, line 50: Nice!!! Have you tried to spawn a network with more than one node with Docker compose?

cmd/eudico-stats/entity.go
```diff
@@ -0,0 +1,139 @@
...
+		SubnetCount: count,
+		Consensus:   output.Consensus,
+		Stake:       output.Subnet.Stake.String(),
+		Status:      output.Subnet.Status,
+		Version:     0,
+	}
+}
+
+func (s *SubnetNode) Equals(o *SubnetNode) bool {
+	return s.SubnetID == o.SubnetID &&
```
adlrocha: Why not doing `s==o` directly?

cmd/eudico-stats/listen.go
```diff
@@ -0,0 +1,379 @@
...
+)
+
+const Timeout = 76587687658765876
+const Confidence = 0
+
+type SubnetStat struct {
+	// The nextEpoch to resync stats with the chain
+	nextEpoch abi.ChainEpoch
+	// The list of subnets currently managed by the current subnet
+	subnets map[address.SubnetID]bool
```
adlrocha, line 29: I don't get this variable. Aren´t you handling already the list of subnets being managed in `EudicoStatsListener`?

cryptoAtwill: This one tracks the children in memory.

cmd/eudico-stats/listen.go
```diff
@@ -0,0 +1,379 @@
...
+// Listen TODO: placeholder for future public invocation differences compared to `listen`
+func (e *EudicoStatsListener) Listen(ctx context.Context, id address.SubnetID) error {
+	return e.listen(ctx, id)
+}
+
+func (e *EudicoStatsListener) listen(
+	ctx context.Context,
+	id address.SubnetID,
+) error {
+	// TODO: ideally there is no need to add read write lock. Keep in view.
```
adlrocha: You may need a lock here, what if the listener accesses the map because you start syncing with a new subnet while you are reading from it to collect some metrics? I'd suggest running with `race` to see if this is the case.

chain/consensus/hierarchical/subnet/submgr/manager.go
cryptoAtwill: I'll remove all the current cross net stats. I'll do it in the event listener.

cmd/eudico-stats/main.go
```diff
@@ -107,29 +139,64 @@ var runCmd = &cli.Command{
...
-					stats.Record(ctx, metrics.TipsetCollectionHeightExpected.M(expectedHeight))
-				}
+			err = eudicoStats.Listen(ctx, address.RootSubnet)
+			if err != nil {
+				log.Errorw("cannot listen to root net")
+				return
 			}
 		}()
 
+		//config := make(map[string]string, 10)
```
adlrocha: clean?

cmd/eudico-stats/listen.go
cryptoAtwill: the list of subnets are for the nodes. This one is to maintain the list of child subnets. Maybe I should give it a differernt name.

cmd/eudico-stats/stats.go
```diff
@@ -0,0 +1,229 @@
...
+//
+//func newSubnetNode(output sca.SubnetOutput, count int) SubnetNode {
+//	return SubnetNode{
+//		SubnetID:    output.Subnet.ID,
+//		SubnetCount: count,
+//		Consensus:   output.Consensus,
+//		Subnet:      output.Subnet,
+//		Version:     0,
+//	}
+//}
```
adlrocha: clean?

cmd/eudico/daemon.go
```diff
@@ -193,6 +199,24 @@ func daemonCmd(overrides node.Option) *cli.Command {
 			// Set the metric to one so it is published to the exporter
 			stats.Record(ctx, metrics.LotusInfo.M(1))
 
+                        // FIXME: Consider moving this to an independent optional process
+                        // (so that is not dependent on the main process).
+			if cctx.Bool("export-metrics") {
+				go func() {
```
adlrocha, line 205: Is this still needed? `eudico-stats`now runs in its own process, right?

cmd/eudico-stats/stats.go
cryptoAtwill: oh yeah, this is the old method I used for reference. I guess it's not much of a use now.

docker-compose.eudico.yaml
```diff
@@ -0,0 +1,104 @@
...
+      - 7687:7687
+    environment:
+      NEO4J_AUTH: neo4j/90d8kOfU*^cR
+
+  neodash:
+    image: nielsdejong/neodash:latest
+    ports:
+      - 5005:5005
+
+#  all-in-one:
```
adlrocha: clean?

cmd/eudico/daemon.go
cryptoAtwill: There might be metrics that cannot be collected by `eudico-stats`, such as response time and stuff like that?

scripts/eudico.sh
```diff
@@ -0,0 +1,81 @@
+#!/usr/bin/env bash
```
adlrocha: What about renaming this script `eudico-stats` or `eudico-stats-example`?

cmd/eudico-stats/entity.go
cryptoAtwill: this does a pointer compare? not the deep compare using pointers?

cmd/eudico-stats/README.md
cryptoAtwill: I will give it a try

PR #234, opened 2022-07-28
Initial reconfiguration mechanism for Mir

### LINKED ISSUE #235: Data race in subnet manager's exchange protocol
### Describe the Bug

Data race: read at `eudico/chain/consensus/hierarchical/subnet/submgr/exchange.go:62`[ref](https://github.com/filecoin-project/eudico/blob/eudico/chain/consensus/hierarchical/subnet/submgr/exchange.go#L62) and write at `eudico/chain/consensus/hierarchical/subnet/submgr/exchange.go:69` [ref](https://github.com/filecoin-project/eudico/blob/eudico/chain/consensus/hierarchical/subnet/submgr/exchange.go#L69)

### Logging Information

```text
==================
WARNING: DATA RACE
Read at 0x00c0051b1350 by goroutine 1318:
  runtime.mapaccess1_faststr()
      /opt/homebrew/Cellar/go/1.18.2/libexec/src/runtime/map_faststr.go:13 +0x41c
  github.com/filecoin-project/lotus/chain/consensus/hierarchical/subnet/submgr.(*Subnet).helloBack()
      /Users/alpha/Projects/eudico/chain/consensus/hierarchical/subnet/submgr/exchange.go:62 +0x1c8
  github.com/filecoin-project/lotus/chain/consensus/hierarchical/subnet/submgr.(*Subnet).handleHelloStream()
      /Users/alpha/Projects/eudico/chain/consensus/hierarchical/subnet/submgr/exchange.go:193 +0x634
  github.com/filecoin-project/lotus/chain/consensus/hierarchical/subnet/submgr.(*Subnet).handleHelloStream-fm()
      <autogenerated>:1 +0x48
  github.com/libp2p/go-libp2p/p2p/host/basic.(*BasicHost).SetStreamHandler.func1()
      /Users/alpha/go/pkg/mod/github.com/libp2p/go-libp2p@v0.19.3/p2p/host/basic/basic_host.go:573 +0x80
  github.com/libp2p/go-libp2p/p2p/host/basic.(*BasicHost).newStreamHandler.func1()
      /Users/alpha/go/pkg/mod/github.com/libp2p/go-libp2p@v0.19.3/p2p/host/basic/basic_host.go:416 +0x70

Previous write at 0x00c0051b1350 by goroutine 1335:
  runtime.mapaccess2_faststr()
      /opt/homebrew/Cellar/go/1.18.2/libexec/src/runtime/map_faststr.go:108 +0x43c
  github.com/filecoin-project/lotus/chain/consensus/hierarchical/subnet/submgr.(*Subnet).helloBack()
      /Users/alpha/Projects/eudico/chain/consensus/hierarchical/subnet/submgr/exchange.go:69 +0x230
  github.com/filecoin-project/lotus/chain/consensus/hierarchical/subnet/submgr.(*Subnet).handleHelloStream()
      /Users/alpha/Projects/eudico/chain/consensus/hierarchical/subnet/submgr/exchange.go:193 +0x634
  github.com/filecoin-project/lotus/chain/consensus/hierarchical/subnet/submgr.(*Subnet).handleHelloStream-fm()
      <autogenerated>:1 +0x48
  github.com/libp2p/go-libp2p/p2p/host/basic.(*BasicHost).SetStreamHandler.func1()
      /Users/alpha/go/pkg/mod/github.com/libp2p/go-libp2p@v0.19.3/p2p/host/basic/basic_host.go:573 +0x80
  github.com/libp2p/go-libp2p/p2p/host/basic.(*BasicHost).newStreamHandler.func1()
      /Users/alpha/go/pkg/mod/github.com/libp2p/go-libp2p@v0.19.3/p2p/host/basic/basic_host.go:416 +0x70

Goroutine 1318 (running) created at:
  github.com/libp2p/go-libp2p/p2p/host/basic.(*BasicHost).newStreamHandler()
      /Users/alpha/go/pkg/mod/github.com/libp2p/go-libp2p@v0.19.3/p2p/host/basic/basic_host.go:416 +0x6f4
  github.com/libp2p/go-libp2p/p2p/host/basic.(*BasicHost).newStreamHandler-fm()
      <autogenerated>:1 +0x48
  github.com/libp2p/go-libp2p/p2p/net/mock.(*peernet).handleNewStream.func1()
      /Users/alpha/go/pkg/mod/github.com/libp2p/go-libp2p@v0.19.3/p2p/net/mock/mock_peernet.go:90 +0x58

Goroutine 1335 (running) created at:
  github.com/libp2p/go-libp2p/p2p/host/basic.(*BasicHost).newStreamHandler()
      /Users/alpha/go/pkg/mod/github.com/libp2p/go-libp2p@v0.19.3/p2p/host/basic/basic_host.go:416 +0x6f4
  github.com/libp2p/go-libp2p/p2p/host/basic.(*BasicHost).newStreamHandler-fm()
      <autogenerated>:1 +0x48
  github.com/libp2p/go-libp2p/p2p/net/mock.(*peernet).handleNewStream.func1()
      /Users/alpha/go/pkg/mod/github.com/libp2p/go-libp2p@v0.19.3/p2p/net/mock/mock_peernet.go:90 +0x58
==================
```

### Repo Steps

_No response_

dnkolegov:
This PR introduces the Mir reconfiguration mechanism.

It also fixes https://github.com/filecoin-project/eudico/issues/235


chain/consensus/README.md
```diff
@@ -1,15 +1,15 @@
...
-5. Add the corresponding CLI commands in `cmd/eudico/$CONSENSUS.go`
-6. Adapt [bad blocks cache](https://github.com/filecoin-project/eudico/blob/0306742e553f6bd6260332b501bb65a5bfc16a76/chain/sync.go#L725) for the consensus protocol if needed.
+1. Register a consensus type constant in `chain/consensus/hierarchical/types.go` and add name, finality, checkpoint period values in the corresponding functions.
+2. Instantiate a consensus and the corresponding miner for a subnet via `New()`, `Weight()`, and `Mine()` functions of `chain/consensus/hierarchical/subnet/consensus/consensus.go`.
+3. Create a new file with implementations of genesis block functions in `chain/consensus/hierarchical/actors/subnet/` and in `chain/consensus/hierarchical/actors/subnet/genesis.go`.
+4. Implement the `Consensus` interface defined in `chain/consensus/iface.go` for the target consensus protocol.
+5. Add the corresponding CLI commands in `cmd/eudico/$CONSENSUS.go`.
+6. Adopt [bad blocks cache](https://github.com/filecoin-project/eudico/blob/0306742e553f6bd6260332b501bb65a5bfc16a76/chain/sync.go#L725) for the consensus protocol if needed.
```
matejpavlovic: I think "adapt" was correct.

chain/consensus/hierarchical/validator.go
```diff
@@ -96,3 +109,61 @@ func splitAndTrimEmpty(s, sep, cutset string) []string {
...
+	return set.Size()
+}
+
+func (set *ValidatorSet) F() int {
+	// TODO: what if n=3 and f=0?
+	return (set.N() - 1) / 3
+}
+
+func (set *ValidatorSet) Majority() int {
+	// TODO: what if n=3 and f=0?
```
matejpavlovic: The TODO can be addressed by using [the same approach as in ISS](https://github.com/filecoin-project/mir/blob/f89f005d1c3544448277d3d666832e95090eef90/pkg/iss/iss.go#L1267-L1284):
```go
func maxFaulty(n int) int {
	// assuming n > 3f:
	//   return max f
	return (n - 1) / 3
}


func strongQuorum(n int) int {
	// assuming n > 3f:
	//   return min q: 2q > n+f
	f := maxFaulty(n)
	return (n+f)/2 + 1
}


func weakQuorum(n int) int {
	// assuming n > 3f:
	//   return min q: q > f
	return maxFaulty(n) + 1
}
```
Except the `n` does not need to be a parameter.

chain/consensus/mir/manager.go
```diff
@@ -54,108 +70,131 @@ func NewManager(ctx context.Context, addr address.Address, api v1api.FullNode) (
...
-	issProtocol, err := iss.New(t.NodeID(mirID), issConfig, newMirLogger(managerLog))
+	netTransport, err := mirlibp2p.NewTransport(h, t.NodeID(mirID), newMirLogger(managerLog))
 	if err != nil {
-		return nil, fmt.Errorf("could not instantiate ISS protocol module: %w", err)
+		return nil, fmt.Errorf("failed to create transport: %w", err)
+	}
+	if err := netTransport.Start(); err != nil {
+		return nil, fmt.Errorf("failed to start transport: %w", err)
 	}
+	netTransport.Connect(ctx, initialMembership)
```
matejpavlovic, line 136: If one of the initial nodes is already down or never starts, will this line block forever?

chain/consensus/mir/manager.go
```diff
@@ -54,108 +70,131 @@ func NewManager(ctx context.Context, addr address.Address, api v1api.FullNode) (
...
 
-	sm := NewStateManager()
+	// Instantiate the ISS protocol module with default configuration.
+
+	issConfig := iss.DefaultConfig(initialMembership)
+	issConfig.ConfigOffset = ConfigOffset
+	issProtocol, err := iss.New(
+		t.NodeID(mirID),
+		issConfig,
+		&commonpb.StateSnapshot{
```
matejpavlovic: The new version of ISS now has a method for returning the initial snapshot (`iss.InitialStateSnapshot`), it's probably cleaner to use that.

chain/consensus/mir/manager.go
```diff
@@ -186,14 +224,40 @@ func (m *Manager) Stop() {
...
 		log.Errorf("Could not close write-ahead log: %s", err)
 	}
 	log.Info("WAL closed")
 
 	m.Net.Stop()
 }
 
+// ReconfigureMirNode reconfigures the Mir node.
+func (m *Manager) ReconfigureMirNode(ctx context.Context, nodes map[t.NodeID]t.NodeAddress) error {
+	log.With("miner", m.MirID).Infof("Reconfiguring a Mir node started")
```
matejpavlovic: None of the statements in this function is blocking. Does it still make sense to log a message both at the start and at the end?

chain/consensus/mir/mine.go
```diff
@@ -33,22 +39,38 @@ func Mine(ctx context.Context, addr address.Address, api v1api.FullNode) error {
...
+
+	lastValidatorSetHash, err := m.LastValidatorSet.Hash()
+	if err != nil {
+		return err
+	}
+
 	for {
+		// Here we use `ctx.Err()` in the beginning of the `for` loop instead of using it in the `select` statement,
+		// because if `ctx` has been closed then `api.ChainHead(ctx)` returns an error,
+		// and we will be in the infinite loop dut to `continue`.
```
matejpavlovic: Typo

chain/consensus/mir/mine.go
```diff
@@ -57,17 +79,70 @@ func Mine(ctx context.Context, addr address.Address, api v1api.FullNode) error {
...
 
 		select {
-		case <-ctx.Done():
-			log.With("epoch", nextEpoch).Debug("Mir miner: context closed")
-			return nil
 		case err := <-mirErrors:
-			return fmt.Errorf("miner consensus error: %w", err)
+			// return fmt.Errorf("miner consensus error: %w", err)
+			//
+			// TODO: This is a temporal solution while we are discussing that issue
```
matejpavlovic: temporal -> temporary

chain/consensus/mir/mine.go
```diff
@@ -57,17 +79,70 @@ func Mine(ctx context.Context, addr address.Address, api v1api.FullNode) error {
...
+
+			// TODO: this is an initial version of reconfiguration mechanism.
+			// In reality, SCA must call us to signal that something has been changed.
+			// For example, two changes may occur between reads and if validators read the state at different times
+			// they could get different val sets.
+
+			// TODO: decide what should we do with environment variable for root net and subnet actor for subnet.
+			// if the variable is empty then we will get warn messages. And it must be empty because otherwise
+			// it will be prioritized on subnet actor config for a subnet.
+			// But to run Mir in rot and in a subnet and we must to use both.
```
matejpavlovic: Not quite sure I understand what that means. Do you mean that an environment variable can override a static configuration? What exactly is the problem?

rot -> root

chain/consensus/mir/state_manager.go
```diff
@@ -0,0 +1,263 @@
...
+	// For each epoch number, stores the corresponding membership.
+	memberships    []map[t.NodeID]t.NodeAddress
+	membershipLock sync.Mutex
+
+	// Channel to send messages to Eudico.
+	ChainNotify chan []Messages
+
+	MirManager *Manager
+
+	reconfigurationVotes map[string]uint64
```
matejpavlovic: The vote counting is leaking memory. That's fine for now, but let's add a TODO here to eventually resolve that.

chain/consensus/mir/state_manager.go
```diff
@@ -0,0 +1,263 @@
...
+
+type Messages []byte
+
+type StateManager struct {
+
+	// The current epoch number.
+	currentEpoch t.EpochNr
+
+	// For each epoch number, stores the corresponding membership.
+	memberships    []map[t.NodeID]t.NodeAddress
```
matejpavlovic: Didn't you want to replace this slice by a map at somepoint...? (Maybe you didn't and I don't remember correctly.)

chain/consensus/mir/state_manager.go
```diff
@@ -0,0 +1,263 @@
...
+
+	reconfigurationVotes map[string]uint64
+}
+
+func NewStateManager(initialMembership map[t.NodeID]t.NodeAddress, m *Manager) *StateManager {
+	// Initialize the membership for the first epochs.
+	// We use configOffset+2 memberships to account for:
+	// - The first epoch (epoch 0)
+	// - The configOffset epochs that already have a fixed membership (epochs 1 to configOffset)
+	// - The membership of the following epoch (configOffset+1) initialized with the same membership,
```
matejpavlovic, line 58: It seems to me that we don't need the extra membership here. It was only used in the old version of the chat app, because the new membership was being progressively constructed (and even there it was actually an ugly solution to keep it directly in the slice).
Would it make sense to have the memberships contain only the agreed-upon memberships in the case of the state manager? Whenever a new membership is agreed upon, it could be just appended here.

chain/consensus/mir/state_manager.go
```diff
@@ -0,0 +1,263 @@
...
+		case ReconfigurationType:
+			err := sm.applyConfigMsg(req.Req)
+			if err != nil {
+				return events.EmptyList(), err
+			}
+		}
+	}
+
+	// Send messages to the Eudico node.
+	sm.ChainNotify <- msgs
```
matejpavlovic: Maybe you can save yourself the trouble of sending the messages through a channel and submit them to Eudico directly from here (the Eudico API seems to be accessible throught the MirManager here anyway).
It might simplify the code.

chain/consensus/mir/state_manager.go
matejpavlovic: The type is `uint64` but it's used only as `int` (converted to it). Can it directly be an `int`?

chain/consensus/mir/state_manager.go
```diff
@@ -0,0 +1,263 @@
...
+	if voted {
+		err = sm.UpdateNextMembership(newValSet)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// applyNewEpoch is applied ?
```
matejpavlovic: The comment doesn't really say much... (I'm not saying there even has to be one, it's pretty clearly an event application function).

chain/consensus/mir/state_manager.go
```diff
@@ -0,0 +1,263 @@
...
+	// The base membership is the last one membership.
+	newMembership := sm.memberships[newEpoch.EpochNr+ConfigOffset]
+
+	// Append a new membership data structure to be modified throughout the new epoch.
+	sm.memberships = append(sm.memberships, newMembership)
+
+	// Update current epoch number.
+	sm.currentEpoch = t.EpochNr(newEpoch.EpochNr)
+
+	err := sm.MirManager.ReconfigureMirNode(context.TODO(), newMembership)
```
matejpavlovic: This doesn't seem like the nicest way to communicate with the MirManager. The context for the reconfiguration is not available here. To use the miner's context, you can communicate the new membership to a channel (similarly as you now do with the transaction batches), and the main MirManager loop would call the `ReconfigureMirNode` method (which currently seems to consists only of connecting to the new nodes). It seems cleaner that way.

chain/consensus/mir/mine.go
```diff
@@ -57,17 +79,70 @@ func Mine(ctx context.Context, addr address.Address, api v1api.FullNode) error {
 
 		// Miner (leader) for an epoch is assigned deterministically using round-robin.
 		// All other validators use the same Miner in the block.
-		epochMiner := sm.Validators[int(base.Height())%len(sm.Validators)].Addr
+		epochMiner := m.GetBlockMiner(base.Height())
```
matejpavlovic: Can the epoch miner be established only just before creating the block? That would be much cleaner, as it is only used there, plus it would be more obviously synchronized with the current state of the membership.

chain/consensus/mir/state_manager.go
```diff
@@ -0,0 +1,263 @@
...
+
+	keys := make([]t.NodeID, len(m))
+	for k := range skeys {
+		keys[k] = t.NodeID(skeys[k])
+	}
+
+	return keys
+}
+
+func copyMap(m map[t.NodeID]t.NodeAddress) map[t.NodeID]t.NodeAddress {
```
matejpavlovic, line 291: you can use `maputil.Copy`

chain/consensus/mir/state_manager.go
```diff
@@ -0,0 +1,263 @@
...
+type Messages []byte
+
+type StateManager struct {
+
+	// The current epoch number.
+	currentEpoch t.EpochNr
+
+	// For each epoch number, stores the corresponding membership.
+	memberships    []map[t.NodeID]t.NodeAddress
+	membershipLock sync.Mutex
```
matejpavlovic: I'm quite convinced that we should not be using the `membershipLock`. The accesses to the `memberships` must be sequential, since they form part of the deterministically updated state. In fact, if they are accessed concurrently, there is some problem with the program logic, as it introduces non-determinism to the state machine. I.e., if it doesn't work without a lock, it will not work with a lock either. The good news is that I think that the access actually is sequential, except potentially for computing the next proposer ID, which can easily be made sequential as well, as proposed in another comment.

matejpavlovic (changes requested):
Looks good in general, I just found some very minor issues.

chain/consensus/README.md
dnkolegov: fixed

chain/consensus/mir/manager.go

chain/consensus/mir/manager.go
dnkolegov: Removed defer's one and changed the log level to Debug

chain/consensus/mir/mine.go
dnkolegov: ok

chain/consensus/mir/mine.go
dnkolegov: ok

chain/consensus/mir/state_manager.go
dnkolegov: fixed

chain/consensus/mir/state_manager.go
dnkolegov: fixed

chain/consensus/mir/state_manager.go
dnkolegov: Eudico supports Go 1.17. The project is not built with features of Go 1.18+

chain/consensus/mir/mine.go
dnkolegov: Fixed the TODO.

chain/consensus/mir/state_manager.go
dnkolegov: Created a TODO

chain/consensus/hierarchical/validator.go
dnkolegov: fixed

chain/consensus/mir/state_manager.go
dnkolegov: Fixed here - https://github.com/filecoin-project/eudico/pull/234/commits/1eed73363a41da77422f722cdf077e73d17b8a99

chain/consensus/mir/state_manager.go
dnkolegov: Yes, you are right. It is possible until the state manager contains access to the Mir manager and the Mir manager has access to the Eudico node API.

But:

- it would extend the logic of Mine function over 2 files
- it is not a function of the state manager (within this design) to create concrete blocks (Filecoin blocks) as we discussed
- it will differ from the style of all other Mine functions

So I would prefer to use the current approach at least for now.

chain/consensus/mir/state_manager.go
dnkolegov: Fixed - https://github.com/filecoin-project/eudico/pull/234/commits/5d32acbac73233852da1d5726be33565778736a8

chain/consensus/mir/mine.go
dnkolegov: Fixed - https://github.com/filecoin-project/eudico/pull/234/commits/a014449298f6902fa97a65f764ed6cfed8a45408

chain/consensus/mir/state_manager.go
dnkolegov: Access is sequential within the Mir. But there is another goroutine - Manager.
Removing the lock leads to this:

```

WARNING: DATA RACE
Read at 0x00c002a6ad50 by goroutine 970:
  github.com/filecoin-project/lotus/chain/consensus/mir.(*StateManager).OrderedValidatorsAddresses()
      /Users/alpha/Projects/eudico/chain/consensus/mir/state_manager.go:206 +0x60
  github.com/filecoin-project/lotus/chain/consensus/mir.(*Manager).GetBlockMiner()
      /Users/alpha/Projects/eudico/chain/consensus/mir/manager.go:245 +0x2c
  github.com/filecoin-project/lotus/chain/consensus/mir.Mine()
      /Users/alpha/Projects/eudico/chain/consensus/mir/mine.go:160 +0x10dc
  github.com/filecoin-project/lotus/chain/consensus/hierarchical/subnet/consensus.Mine()
      /Users/alpha/Projects/eudico/chain/consensus/hierarchical/subnet/consensus/consensus.go:81 +0x184
  github.com/filecoin-project/lotus/chain/consensus/hierarchical/subnet/submgr.(*Subnet).mine.func2()
      /Users/alpha/Projects/eudico/chain/consensus/hierarchical/subnet/submgr/subnet.go:294 +0x9c

Previous write at 0x00c002a6ad50 by goroutine 796:
  github.com/filecoin-project/lotus/chain/consensus/mir.(*StateManager).applyNewEpoch()
      /Users/alpha/Projects/eudico/chain/consensus/mir/state_manager.go:138 +0x314
  github.com/filecoin-project/lotus/chain/consensus/mir.(*StateManager).ApplyEvent()
      /Users/alpha/Projects/eudico/chain/consensus/mir/state_manager.go:74 +0x9c
  github.com/filecoin-project/lotus/chain/consensus/mir.(*StateManager).ApplyEvent-fm()
      <autogenerated>:1 +0x40
  github.com/filecoin-project/mir/pkg/modules.ApplyEventsSequentially()
```

chain/consensus/mir/manager.go
dnkolegov: Now this line blocks forever but the new libp2p transport PR will fix it by returning an error after `retryAttempts * connectingTimeout` seconds. 

See https://github.com/filecoin-project/mir/pull/190/commits/405d56a06d47fe387621dabe27950c7f5f6c299b

chain/consensus/mir/state_manager.go
dnkolegov: Can you elaborate more on how to fix this without locking?

dnkolegov: Implemented this by another way - https://github.com/filecoin-project/eudico/pull/234/commits/d1bf8110ea7d0d295e33c5726e3e5b6f9a1ad5c5

chain/consensus/mir/state_manager.go
dnkolegov: I implemented removing memberships for the old epochs. Did it fix this comment?

matejpavlovic: What I meant was slightly different, but it's ok the way it is now.

chain/consensus/mir/state_manager.go
matejpavlovic: That's true, but this part of the code is Mir-specific, and Mir needs Go 1.18+ anyway, so not using generics here doesn't seem of any benefit.

That said, it's a very minor, mostly aesthetic issue, so feel free to keep this separate implementation if you prefer and resolve this comment yourself.

chain/consensus/mir/state_manager.go
matejpavlovic: Fair enough.

matejpavlovic (approved):
Good job, thanks Denis! 👍

chain/consensus/mir/state_manager.go
dnkolegov: I cannot do that without changing Go version for eudico/Lotus
```
chain/consensus/mir/state_manager.go:210:35: implicit function instantiation requires go1.18
or later (-lang was set to go1.17; check go.mod)
```

matejpavlovic: Ah ok, then of course let's keep it.

chain/consensus/hierarchical/actors/subnet/subnet_actor.go
```diff
@@ -203,8 +204,13 @@ func (a SubnetActor) Leave(rt runtime.Runtime, _ *abi.EmptyValue) *abi.EmptyValu
 	priorBalance := rt.CurrentBalance()
 	var retFunds abi.TokenAmount
 	rt.StateTransaction(&st, func() {
-		// Remove stake from stake balanace table.
+		// Remove stake from stake balance table.
 		retFunds = st.rmStake(rt, sourceAddr, stakes, minerStake)
+
+		// Remove the validator to signal to other validators on the subnet to run reconfiguration.
+		if st.ValidatorSet != nil {
```
adlrocha, line 211: We should remove `MinerList` in favor of `ValidatorSet` and include all the information in `ValidatorSet` (so you don't have to do this check). In any case, not worth changing it here because soon we will deprecate these actors, and we'll only have the FVM actors, so we need to implement this functionality in the Rust actor. This approach LGTM.

chain/consensus/mir/consensus.go
```diff
@@ -36,9 +36,13 @@ import (
 )
 
 const (
-	MaxHeightDrift = 5
-	SubmitInterval = 5000 * time.Millisecond
-	ValidatorsEnv  = "EUDICO_MIR_VALIDATORS"
+	SubmitInterval          = 5000 * time.Millisecond
```
adlrocha: Should this become a build parameter?

chain/consensus/mir/manager.go
```diff
@@ -54,108 +69,178 @@ func NewManager(ctx context.Context, addr address.Address, api v1api.FullNode) (
...
+		&simplemempool.ModuleConfig{
+			Self:   "mempool",
+			Hasher: "hasher",
+		},
+		&simplemempool.ModuleParams{
+			MaxTransactionsInBatch: 10,
+		},
+	)
+
+	mscFactory := factorymodule.New(
```
adlrocha, line 169: Ok, I didn't get a thing of this, I may need to learn once and for all the Mir base code. If @matejpavlovic doesn't see anything weird here, LGTM :)

chain/consensus/mir/manager.go
```diff
@@ -186,20 +270,34 @@ func (m *Manager) Stop() {
...
 
+// ReconfigureMirNode reconfigures the Mir node.
+func (m *Manager) ReconfigureMirNode(ctx context.Context, nodes map[t.NodeID]t.NodeAddress) error {
+	log.With("miner", m.MirID).Debug("Reconfiguring a Mir node")
+
+	if len(nodes) == 0 {
+		return fmt.Errorf("empty validator set")
+	}
+
+	go m.Net.Connect(ctx, nodes)
```
adlrocha, line 289: Just to understand what is happening under the hood. We connect to all nodes (in case there is one we are not connected to in the new set), and we garbage collect the connections of the nodes that left, right? 

Why do we spawn different routines for this? Can this lead to a data race if Mir tries to send a message to a node before we have successfully connected to it (maybe there is an internal lock that handles this and makes this operation concurrent-friendly, but I wanted to double-check just in case).

chain/consensus/mir/manager.go
```diff
@@ -271,14 +369,25 @@ func (m *Manager) GetMessages(batch []Tx) (msgs []*types.SignedMessage, crossMsg
...
-func (m *Manager) GetRequests(msgs []*types.SignedMessage, crossMsgs []*types.UnverifiedCrossMsg) (
+func (m *Manager) TransportRequests(msgs []*types.SignedMessage, crossMsgs []*types.UnverifiedCrossMsg) (
 	requests []*mirproto.Request,
 ) {
 	requests = append(requests, m.batchSignedMessages(msgs)...)
 	requests = append(requests, m.batchCrossMessages(crossMsgs)...)
 	return
 }
 
+func (m *Manager) ReconfigurationRequest(payload []byte) *mirproto.Request {
```
adlrocha, line 380: @matejpavlovic, out of a curiosity, do we reconfigure as soon as we see a change in the validator set, or is there a "reconfiguration epoch" where we make the joining/leaving of a node effective? I am starting to think the latter can prevent potential attacks, but not sure...

chain/consensus/mir/mine.go
```diff
@@ -1,76 +1,159 @@
...
 //
-// Mine implements the following algorithm:
+// Mine implements the following main algorithm:
 // 1. Retrieve messages and cross-messages from the mempool.
-//    Note, that messages can be added into mempool via the libp2p mechanism and the CLI.
+//    Note, that messages can be added into mempool via the libp2p and CLI mechanism.
 // 2. Send messages and cross messages to the Mir node through the request pool implementing FIFO.
 // 3. Receive ordered messages from the Mir node and parse them.
-// 4. Create the next Filecoin block. Note, only a leader Eudico node, chosen by round-robin election, creates a block.
+// 4. Create the next Filecoin block. Note, only the leader Eudico node, chosen by round-robin election,
```
adlrocha, line 26: We should definitely change this and make a Reward interface that allows, according to the consensus algorithm, to perform a different reward distribution. I guess in Mir's case we should distribute the rewards between all the validators.

I'll add an issue to track this.

chain/consensus/mir/mine.go
```diff
@@ -1,76 +1,159 @@
...
+//    creates a block and gets a reward. This simulation is used to conform the Filecoin consensus.
 // 5. Sync this block without sending it over the libp2p network.
```
adlrocha, line 28: What if a node is a full node but not a validator? How it is informed about new blocks if we don't broadcast them?

chain/consensus/mir/manager.go
adlrocha: Question answered :) https://github.com/filecoin-project/eudico/blob/2ab9450496f49e0724cebd18fc16117759aa8544/chain/consensus/mir/mine.go#L57

chain/consensus/mir/mine.go
```diff
@@ -1,76 +1,159 @@
...
-			return nil
+
 		case err := <-mirErrors:
-			return fmt.Errorf("miner consensus error: %w", err)
-		case batch := <-mirHead:
-			msgs, crossMsgs := sm.GetMessages(batch)
+			// return fmt.Errorf("miner consensus error: %w", err)
+			//
+			// TODO: This is a temporary solution while we are discussing that issue
+			// https://filecoinproject.slack.com/archives/C03C77HN3AS/p1660330971306019
```
adlrocha, line 83: Did we reach a conclusion?

chain/consensus/mir/mine.go
```diff
@@ -1,76 +1,159 @@
...
+		case <-reconfigure.C:
+			// Reconfiguration is not used in the rootnet.
+			if m.SubnetID == address.RootSubnet {
+				continue
+			}
+			//
+			// Send a reconfiguration transaction if the validator set in the actor has been changed.
+			//
+
+			// NOTE: You must unset the environment variable in tests if you use Mir in the rootnet and in a subnet.
```
adlrocha, line 101: We'll need to add a membership actor to support also proof-of-stake consensus in rootnets, so we may have a solution to this soon.

chain/consensus/mir/mine.go
```diff
@@ -1,76 +1,159 @@
...
+			// So if you don't unset the variable after instantiation Mir in the rootnet
+			// the subnet Mir miner cannot get membership.
+			// The environment variable must be empty because otherwise it will be prioritized for a subnet.
+			newValidatorSet, err := getSubnetValidators(ctx, m.SubnetID, api)
+			if err != nil {
+				log.With("epoch", nextEpoch).Warnf("failed to get subnet validators: %v", err)
+				continue
+			}
+
+			newValidatorSetHash, err := newValidatorSet.Hash()
```
adlrocha: hahaha this is an interesting way of checking if there's a diff. Do you think that marshaling and computing the hash is more performant than looping through the map to check the diff'ed elements?

chain/consensus/mir/state_manager.go
```diff
@@ -0,0 +1,309 @@
...
+	Validators []t.NodeID
+}
+
+type StateManager struct {
+
+	// The current epoch number.
+	currentEpoch t.EpochNr
+
+	// For each epoch number, stores the corresponding membership.
+	memberships map[t.EpochNr]map[t.NodeID]t.NodeAddress
```
adlrocha, line 39: Why do we need to store the history of memberships for every epoch?

adlrocha: I see, you use it just to keep the latest two epochs, you don't keep the whole history.

chain/consensus/mir/state_manager.go
```diff
@@ -0,0 +1,309 @@
...
+			},
+		)), nil
+	default:
+		return nil, fmt.Errorf("unknown availability certificate type: %T", deliver.Cert.Type)
+	}
+}
+
+// applyProvideTransactions applies transactions received from the availability layer to the app state.
+// In our case, it simply extends the message history
+// by appending the payload of each received request as a new chat message.
```
adlrocha: typo?
```suggestion
// by appending the payload of each received request as a new message.
```

itests/kit/tools_eudico.go
```diff
@@ -91,67 +100,117 @@ func WaitForBalance(ctx context.Context, addr addr.Address, balance uint64, api
...
+func SubnetHeightCheckForBlocks(ctx context.Context, n int, subnetAddr addr.SubnetID, api napi.FullNode) error {
+	heads, err := getSubnetChainNotify(ctx, subnetAddr, api)
 	if err != nil {
 		return err
 	}
-	height := head.Height()
-	if height != currHeight {
-		return xerrors.New("wrong initial block height")
+
+	select {
```
adlrocha, line 193: Why is this first select block needed?

chain/consensus/mir/mine.go
```diff
@@ -1,76 +1,159 @@
...
+			err = newValidatorSet.MarshalCBOR(&payload)
+			if err != nil {
+				log.With("epoch", nextEpoch).Warnf("failed to marshal validators: %v", err)
+				continue
+			}
+			m.SubmitRequests(ctx, []*mirproto.Request{
+				m.ReconfigurationRequest(payload.Bytes())},
+			)
+
+		case batch := <-m.App.ChainNotify:
```
adlrocha, line 134: Nice approach to decouple the Mir and mining logic :raised_hands:

adlrocha (approved):
Amazing job, @dnkolegov, @matejpavlovic. I left a few comments just for sanity-check and personal understanding, but I don't think there are any major blockers, feel free to merge if my comments don't rise any big issue. :pray:

adlrocha:
@dnkolegov, do you mind looking [at this](https://github.com/filecoin-project/eudico/issues/236#issuecomment-1227034717) before merging? We can look at it together if needed. Thanks :pray:

chain/consensus/mir/manager.go
dnkolegov: Actually, that's @matejpavlovic 's reference code for using the availability layer.

chain/consensus/mir/mine.go
dnkolegov: Added - https://github.com/filecoin-project/eudico/issues/239

chain/consensus/mir/state_manager.go
dnkolegov: Yes, there is a garbage collection for old memberships.

itests/kit/tools_eudico.go
dnkolegov: Without that first select, currHead may be equal to newHead. That is because we could read them both at the same height.
If we call SubnetChainHead twice we can get the same head. But if we call it after reading from the unbuffered channel we have a guarantee that the blocks will be different.

chain/consensus/mir/mine.go
dnkolegov: Got it! Make sense.

chain/consensus/mir/mine.go
dnkolegov: Probably we should use panicking in Mir's Mine function. Otherwise, we need to redesign the subnet.

chain/consensus/mir/manager.go
dnkolegov: We reconfigure as soon as we get f+1 votes on the reconfiguration request.

`reconfigure ` ticker is how often we check the validator set via the subnet actor. Then we send a special transaction with the new validator set. If it gets f+1 votes then it will be applied in +2 epochs.

chain/consensus/mir/manager.go
dnkolegov: > Just to understand what is happening under the hood. We connect to all nodes (in case there is one we are not connected to in the new set), and we garbage collect the connections of the nodes that are left, right?

That's right.

> Why do we spawn different routines for this?

Mir's libp2p transport can try to establish connections with a node very long time. To not block this we spawn a new goroutine. We believe that adding a node is a rare operation.

@matejpavlovic can elaborate more about possible data race in Mir

chain/consensus/mir/consensus.go
dnkolegov: SubmitInterval is not used. Removed

chain/consensus/hierarchical/actors/subnet/subnet_actor.go
dnkolegov: ok

chain/consensus/mir/mine.go
dnkolegov: Implemented your suggestion - https://github.com/filecoin-project/eudico/pull/234/commits/12fceca092df2fbd3caedb2714c3aff4c64f11da

chain/consensus/mir/mine.go
dnkolegov: I thought that the Sync protocol can be used for that.
If not then we must address this in the next iteration.

chain/consensus/mir/manager.go
adlrocha: @matejpavlovic, can I invite you to a coffee/beer in exchange for a quick overview of the Mir base code? I feel I can be significantly more helpful and less annoying in my reviews if I steal an hour of your time for this. Thanks.

chain/consensus/mir/manager.go
adlrocha: That makes sense as long as Mir message sender is blocked for the corresponding peer until the connection is established. Let's wait for @matejpavlovic input.

chain/consensus/mir/manager.go
adlrocha: LGTM. This means that effectively there is no reconfiguration until this ticker checks the validator set, right? (this ticker is basically the epoch I was referring to, we don't want a peer leaving and joining to force continuous reconfigurations DoS'ing the protocol).

chain/consensus/mir/mine.go
adlrocha: The Sync protocol handles that as long as it is notified that there is a new block (our current view is below the heaviest height), which I don't think is the case in the current implementations. Have you tried to sync a node that is not a validator with Mir?

I'm fine addressing this (if @matejpavlovic is OK with it) in the next PR, but let's track it in an issue if you don't mind so we don't forget :) )

chain/consensus/mir/mine.go
adlrocha: For now, LGTM.

chain/consensus/mir/mine.go
adlrocha: I know, but I was thinking about doing it in some other way :see_no_evil: by iterating the two maps and checking what was different, not using a hash, that is why it surprised me. How big we thing this maps can be?

itests/kit/tools_eudico.go
adlrocha: Gotcha! Makes sense.

chain/consensus/mir/manager.go
dnkolegov: > This means that effectively there is no reconfiguration until this ticker checks the validator set, right?

That's right.

> we don't want a peer leaving and joining to force continuous reconfigurations DoS'ing the protocol

We probably do not want this. But now any changes in the subnet actor detected by the Mine function trigger reconfiguration in Mir.

chain/consensus/mir/mine.go
dnkolegov: We don't use maps. ValidatorSet is a slice.

> How big we thing this maps can be?

it depends on subnets.

dnkolegov:
> @dnkolegov, do you mind looking at https://github.com/filecoin-project/eudico/issues/236#issuecomment-1227034717 before merging? We can look at it together if needed. Thanks 🙏

@adlrocha I ran tests 3 times via CI. No those data races so far.

chain/consensus/mir/mine.go
adlrocha: Iterate the lists (sorry, typo).

adlrocha:
Beautiful, @dnkolegov Then it looks like its fixed. We need to handle the redundancy between the subnet lock and the manager lock, but I will add an issue and we can add it in a future PR. Let's wait for @matejpavlovic's on the networking race, and feel free to merge if it not a big issue.

chain/consensus/mir/manager.go
matejpavlovic: Of course @adlrocha , jsut send me an invite for any time that is free in my calendar and I'll give you a walk-through.
There have been significant updates to Mir recently and they are not yet reflected in the documentation, so for new features (like this factory module), the current docs are not helpful at all. That's on my todo-list too :-)

Regarding these concrete lines, indeed I wrote them, and yes they're ugly. 😬 They do what they should, but I agree that some cleanup and more thorough comments are missing...

chain/consensus/mir/manager.go
matejpavlovic: Yes indeed, this actually is an issue, good catch. A node could try to send a message to a newly joined node that the Net module hasn't even tried to connect to yet, resulting in an error (which in the current implementation would even lead to a crash). In this concrete case it is pretty unlikely, but still possible.

The Net module needs to be aware of all the node IDs it can be asked to send messages to before any send event. To do that, maybe we could have the `Connect` method not run in a separate goroutine (i.e. just call `m.Net.Connect(ctx, nodes)`, without the `go`), but change its internal implementation to a non-blocking one that creates internal message buffers synchronously, but builds the network connections in the background.

Moreover, we should be more graceful about sending messages to unknown nodes. Simply dropping the message (and maybe logging an error) would be the right thing to do in my opinion.

matejpavlovic: I created an issue in Mir for that.

chain/consensus/mir/manager.go
matejpavlovic: In fact, it's not quite that.
The ticker only determines the polling interval, at which the node checks the SA state whether there were any changes in the validator set. Then everybody "votes" on these changes (not really "changes", but rather "new validator set") using the subnet's consensus protocol, and once there is enough votes for the same new configuration, the new configuration is only **scheduled** to be applied at the next reconfiguration occasion (when the `NewConfig` event is triggered in response to `NewEpoch`). If the configuration is changed multiple times in fast succession, then only the most recent one is actually performed.

chain/consensus/mir/mine.go
dnkolegov: Yes, now it works by iterating the slice.

chain/consensus/mir/manager.go
adlrocha: > The Net module needs to be aware of all the node IDs it can be asked to send messages to before any send event. To do that, maybe we could have the Connect method not run in a separate goroutine (i.e. just call m.Net.Connect(ctx, nodes), without the go), but change its internal implementation to a non-blocking one that creates internal message buffers synchronously, but builds the network connections in the background.

This would be a good fix in the current PR, and I think is a good idea for Mir to have a sanity-check as a library. Thanks!

chain/consensus/mir/mine.go
adlrocha: @matejpavlovic, we are only missing your input here, and LGTM to merge. Thanks!

matejpavlovic: At this stage, when using Mir, there is no such thing as a non-validator node yet. Every node in a mir subnet is a validator. Adding support for non-validating nodes has been discussed as a future extension, but it's not on our plate now.

chain/consensus/mir/mine.go
matejpavlovic: This is such a rare event (once a timer ticker) that I'd be extremely surprised if it had any impact whatsoever on the performance, regardless of the method used. I wouldn't consider performance to be an issue here (unless extremely wasteful) and prioritize the easiest-to-understand solution (both hashing and iterating seem fine to me, I guess it's a matter of personal taste).

chain/consensus/mir/manager.go
matejpavlovic: Actually, there might be a little liveness issue with this approach... Under constant and high churn, the nodes might actually never see the same configuration. We only get liveness guarantees here under the assumption that eventually there will be a period of time without changes to the membership that is long enough for a quorum of correct nodes to observe the same configuration.
I think we can live with this assumption for the time being, I'll create an issue for this and we need to come back to it and implement something more sophisticated to guarantee progress even under constant churn.

matejpavlovic:
Sorry for the delay, responded to all comments now @dnkolegov and @adlrocha .

chain/consensus/platform/platform.go
```diff
@@ -0,0 +1,54 @@
+package platform
```
adlrocha, line 1: Out of a curiosity, why platform? :)

chain/consensus/hierarchical/subnet/submgr/manager.go
```diff
@@ -389,6 +389,9 @@ func (s *Service) GetSubnetState(ctx context.Context, id address.SubnetID, actor
 }
 
 func (s *Service) AddSubnet(ctx context.Context, params *hierarchical.SubnetParams) (address.Address, error) {
+	s.lk.Lock()
```
adlrocha, line 392: Do you mind removing the TODOs in the rest of the subnet manager functions where we use `s.lk`? I don't think we need to re-thing the locking strategy, at least for now.

itests/eudico_hc_test.go
```diff
@@ -783,36 +765,19 @@ func (ts *eudicoSubnetSuite) testDataRaces(t *testing.T) {
 		}
 	}()
 
-	go func() {
```
adlrocha: Why did you decide to go from 3 nodes to 2 nodes in this test? (I guess that for readibility?)

adlrocha:
Ok, this was a tough one, but I think after removing those TODOs we are ready to merge, @dnkolegov. Amazing job leading this one! :pray:

chain/consensus/mir/manager.go
matejpavlovic: The ugliness of the code (to a large extent) fixed by https://github.com/filecoin-project/mir/pull/205

adlrocha: Amazing!!! Thank you, @matejpavlovic.

itests/eudico_hc_test.go
dnkolegov: Those 2 nodes are nodes for the rootnet and the primary subnet. I beleive there is no much sense to start 3 nodes or even 2.
So the setup is as follows:
- 2 nodes in the rootnet with dummy consensus
- 2 nodes in the subnet with Mir
- 4 side subnets with 1 node in each with PoW, Dummy, Delegated, PoW to provide concurrent access to subnet APIs

chain/consensus/hierarchical/subnet/submgr/manager.go
dnkolegov: removed

chain/consensus/platform/platform.go
dnkolegov: That's a good name I learned from ArdanLabs training ;-)

chain/consensus/mir/mine.go
dnkolegov: @adlrocha 
> Have you tried to sync a node that is not a validator with Mir?

A new node joining a subnet is syncing with current nodes. Then it starts producing new blocks by yourself (if the node is miner). If it is not a miner then it stops syncing.

PR #238, opened 2022-08-24
Parametrize PoW to make tests faster

dnkolegov:
The goal of this PR is to clean code style and parametrize the implementation of PoW.

Below you can see the timestamps of blocks mined by PoW on one node in our HC tests.

```
2022-08-24T10:52:06.422Z
2022-08-24T10:52:14.232Z
2022-08-24T10:52:40.816Z
2022-08-24T10:53:17.329Z
2022-08-24T10:53:36.690Z
2022-08-24T10:53:46.726Z
```

It is very inefficient to spend 30-40 seconds to mine a block in tests. 
Because of that PoW based tests often fail.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -376,7 +376,7 @@ jobs:
   build-macos:
     description: build darwin lotus binary
     macos:
-      xcode: "12.5.0"
+      xcode: "12.5.1"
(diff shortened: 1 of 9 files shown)


adlrocha (approved):
LGTM.

PR #250, opened 2022-08-31
Support Mir's event recorder

dnkolegov:
This PR:

- updates Mir version
- introduces an environment variable `EUDICO_INTERCEPTOR_OUTPUT` to enable Mir's event recorder (interceptor)
- modifies Circle CI config to be able to download event logs
- removes Tendermint test files


chain/consensus/mir/manager.go
```diff
@@ -143,6 +153,19 @@ func NewManager(ctx context.Context, addr address.Address, api v1api.FullNode) (
 		return nil, fmt.Errorf("failed to create crypto manager: %w", err)
 	}
 
+	// Instantiate an interceptor.
+
+	var interceptor *eventlog.Recorder
+
+	interceptorOutput := os.Getenv(InterceptorOutputEnv)
```
adlrocha, line 160: What is the interceptor? Should this only be started in "debugging mode" to avoid unnecessary overheads?

adlrocha: Nevermind, I see below that this is the case already :)

adlrocha (approved):
LGTM. Merging time :)

chain/consensus/mir/manager.go
dnkolegov: Yes, I also added that to the PR description.
I was trying to get an interceptor from the context but it is changed when Subnet is started, so I decided to use environment variables

PR #251, opened 2022-09-01
Switch to Go 1.18

dnkolegov:
This PR switches to Go 1.18 following Lotus.


adlrocha (approved):
Looks like it wasn't hard. I have a slight concern with the compatibility with the builder net branch that I keep syncing with `experimental/fvm-m2`, but let's worry about that later.

PR #253, opened 2022-09-13
Eudico and Mir mempools pipeline

dnkolegov:
This PR integrates Eudico and Mir mempools with each other.


chain/consensus/mir/mempool/handlers/batches.go
```diff
@@ -40,10 +40,20 @@ func IncludeBatchCreation(
...
 
+		submitChan := make(chan []*requestpb.Request)
+
+		state.DescriptorChan <- types.Descriptor{
+			Limit:      0,
+			SubmitChan: submitChan,
+		}
+
+		receivedRequests := <-submitChan
+		mpdsl.RequestTransactionIDs(m, mc.Self, receivedRequests, &requestTxIDsContext{receivedRequests})
```
matejpavlovic: Ok I was probably wrong about only having to change this function. I didn't realize that the transaction IDs still must be computed. This implementation only gets the transactions from outside and has their IDs computed, but does not include them in the batch that is emitted as an event. In a strange way it probably works, just delayed, but this is not what I had in mind. What I meant is to directly emit a batch with the transactions received through the channel, without any buffering. For that, there needs to be more change to the module implementation.

matejpavlovic (changes requested):

chain/consensus/mir/mempool/handlers/batches.go
dnkolegov: It seems to me this is a [feature](https://github.com/filecoin-project/mir/blob/8e480ade52d4d62cc762682cfbd6c2a0e25d4c77/pkg/availability/multisigcollector/internal/parts/batchreconstruction/batchreconstruction.go#L96.).

I can delete calling of `mpdsl.RequestTransactionIDs` and implement hashing without sending an event but it will be also called later in batch reconstruction that is a not part of mempool.

Let's discuss in sync

chain/consensus/mir/mempool/handlers/batches.go
```diff
@@ -21,54 +21,37 @@ func IncludeBatchCreation(
...
-		var txIDs []t.TxID
 		var txs []*requestpb.Request
-		batchSize := 0
-
 		txCount := 0
-		for _, txID := range state.NewTxIDs {
-			tx := state.TxByID[txID]
+		for i := range txIDs {
+			tx := context.txs[i]
+			commonState.TxByID[txIDs[i]] = tx
```
matejpavlovic: Is there a reason why we still save the transaction in the commonState buffer? It would be nice if this buffer didn't exist at all and the transactions were directly added in the new batch and immediately forgotten.

chain/consensus/mir/mempool/handlers/batches.go
```diff
@@ -21,54 +21,37 @@ func IncludeBatchCreation(
...
 
 			// TODO: add other limitations (if any) here.
 			if txCount == params.MaxTransactionsInBatch {
```
matejpavlovic: This check should probably not be enforced here any more. The handler should just take all transactions it receives and put them in a batch. The limit should be enforced when the transactions are fetched from the Eudico mempool.

chain/consensus/mir/mempool/handlers/batches.go
dnkolegov: Fixed

chain/consensus/mir/mempool/handlers/batches.go
dnkolegov: Fixed

matejpavlovic (approved):
Very good! I think we can merge this as the first round, and continue working on further improvements in a different PR.
Thanks!

PR #255, opened 2022-10-04
Add GC for votes in reconfiguration mechanism

dnkolegov:


chain/consensus/mir/state_manager.go
```diff
@@ -202,35 +201,43 @@ func (sm *StateManager) applyNewEpoch(newEpoch *eventpb.NewEpoch) (*events.Event
...
+	delete(sm.reconfigurationVotes, oldEpoch)
 
 	sm.NewMembership <- newMembership
 
 	// Notify ISS about the new membership.
 	return events.ListOf(events.NewConfig("iss", newMembership)), nil
 }
 
-func (sm *StateManager) UpdateNextMembership(valSet *hierarchical.ValidatorSet) error {
+func (sm *StateManager) UpdateNextMembership(valSet *hierarchical.ValidatorSet, valSetHash string) error {
```
matejpavlovic: You seem to have added a parameter to a function, but you never use it inside... Is that intended?

matejpavlovic (approved):
LGTM, modulo one small comment.

chain/consensus/mir/state_manager.go
dnkolegov: You are right! I forget to remove it.

